PR #8, opened 2020-07-08
ci: update github actions, add publish workflow for version tags

tommilligan:
Also adds caching for rust dependencies, which should speed up build times.

This will definitely not work until the following secrets are populated by an admin (@mcobzarenco):
- `GCP_SA_KEY`, a GCP Service Account Key which has write access to the `gs://reinfer-public` bucket
- `CRATES_IO_TOKEN`, a crates.io token with publish permissions to the `reinfer-cli` and `reinfer-client` crates.

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+set -ex
+
+cargo check --locked --all-targets
+cargo build --locked --all-targets
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,10 +1,8 @@
-name: Build
+name: Publish
 
 on:
   push:
-    branches: [master]
-  pull_request:
-    branches: [master]
+    tags: v*
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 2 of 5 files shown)


scripts/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -ex
+
+cargo check --all-targets --all-features --locked
+cargo build --locked --all-targets --all-features
```
mcobzarenco, line 6: Something I ran into before, [suprisingly](https://github.com/rust-lang/cargo/issues/6669) `--all-targets` doesn't run doc tests, whereas `cargo test` would. I don't think we have any examples in the docs, but may be worth adding `cargo test --docs ...` too so we don't forget.

Also, suuupper nit. the arguments could be in same order

tommilligan: Good point - yes, I feel like I've been annoyed about that before.

Have un-nitted the arguments, and removed some duplicate/redundant flags (like, we don't need --all-features because we have no features at the moment).

PR #15, opened 2020-07-30
task: add take-more-labels

tommilligan:
Implements Will's request to move labels from a source dataset to a target dataset for **only** reviewed comments in the target dataset with a specific label.

@annesylvie for context on the actual job and label logic, @mcobzarenco for adding the `task` subcommand and the `--experimental` global flag.

Movement of labels follows the following logic:

take_more_labels


tommilligan:
ping both of you guys, low priority but it's been open a while now

api/src/resources/comment.rs
```diff
@@ -498,7 +512,7 @@ pub struct EntityKind(pub String);
 impl FromStr for EntityKind {
     type Err = Error;
 
-    fn from_str(string: &str) -> Result<Self, Error> {
+    fn from_str(string: &str) -> Result<Self> {
```
annesylvie, line 515: Why is this no longer with an `Error`?

api/src/resources/comment.rs
```diff
@@ -237,7 +243,7 @@ const STRING_PROPERTY_PREFIX: &str = "string:";
 const NUMBER_PROPERTY_PREFIX: &str = "number:";
 
 impl Serialize for PropertyMap {
-    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+    fn serialize<S: Serializer>(&self, serializer: S) -> StdResult<S::Ok, S::Error> {
```
annesylvie, line 246: What is the difference between `Result` and `StdResult`?

cli/src/commands/task/take_more_labels.rs
```diff
@@ -0,0 +1,258 @@
...
+/// Collect target comment ids and labellings, which contain the given assigend label.
+fn get_labellings_with_assigned_label(
+    annotated_comments: Vec<AnnotatedComment>,
+    assigned_label_name: &LabelName,
+) -> HashMap<Uid, Labelling> {
+    let mut target_labellings_to_update = HashMap::new();
+    // Consume the comments here as we break them up into parts for reuse
+    for annotated_comment in annotated_comments.into_iter() {
+        let AnnotatedComment {
+            comment, labelling, ..
```
annesylvie, line 151: Interesting `..` notation, not sure what it does? I guess it unpacks.

cli/src/commands/task/take_more_labels.rs
```diff
@@ -0,0 +1,258 @@
...
+            }
+
+            let _progress = make_progress()?;
+        }
+    }
+    Ok(())
+}
+
+/// Collect target comment ids and labellings, which contain the given assigend label.
+fn get_labellings_with_assigned_label(
```
annesylvie, line 143: This function name is a bit weird, isn't it really `map_comments_to_assigned_labels`?

cli/src/commands/task/take_more_labels.rs
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+    target_labellings_to_update
+}
+
+/// Merge the old and new assigned labels together.
+///
+/// Transferred labels will be nested under `transferred_label_parent` unless
+/// they already begin with that parent label.
+fn merge_assigned_labels(
```
annesylvie, line 172: I'd like to chat about this, I have a few questions about this function but it would be easier to talk.

api/src/resources/comment.rs
tommilligan: At the top of the file, bringing
```rust
crate::errors::{Error, Result};
```
into scope shadows the default meaning of `Result` from the standard library. So it is now defined as
```rust
pub type Result<T> = StdResult<T, Error>;
```

api/src/resources/comment.rs
tommilligan: Again, we're defining `Result` as our custom type alias. And we import the standard library's `Result' under a different name:
```rust
use std::result::Result as StdResult;
```

cli/src/commands/task/take_more_labels.rs
tommilligan: Yes, it's saying 'there are other fields in this `AnnotatedComment` and I don't care about them'. Otherwise you would get an error saying there are fields that we have not assigned to a variable.

annesylvie (approved):
LGTM!

mcobzarenco (changes requested):
I am not opposed to adding tasks for one offs or more experimental stuff in the CLI. I'd change how we do it a bit: 

1. This is somewhat reminiscent of a `demo` feature flag under which we had an IMAP integration and a few other things -- my preference would be to make it a feature flag `experimental` that is disabled by default. In that case, you don't need the `experimental` cmd line flag any more, e.g. compiling with the feature flag would give you access to an extra subcommand. Disabling it should remove the cmd line arguments and I'd be up for grouping all extensions under a new command. This way we don't have to expose these tasks to users + we can include some dependencies that are only required for tasks (e.g. the imap example above). I'd probably bring that back from the dead if you added a feature flag.
1. I think we could choose an alternative name to `task` that is less committal to what's experimental, e.g. something like `beta`, `alpha`,  `unstable`, `experimental`? I guess I like `alpha`, but :shrug:  So something like `imap` would fit there naturally, e.g. `re alpha take-more-labels` or `re alpha imap`?

mcobzarenco:
I commented on how I'd structure this, I haven't reviewed the code -- @davidhewitt do you think you could check it out?

tommilligan:
@davidhewitt this is ready for review with conditional compilation for the `alpha` subcommand. Anne-Sylvie has reviewed the actual task logic, so I think this is just around the subcommand structure etc.

The one question I have around this is how to manage testing/checking these features in CI. Obviously I can duplicate a bunch of `cargo clippy` and `cargo test` features in the `check` script to add the `--all-features` flag, but is there a better way you know of?

cli/src/commands/alpha/take_more_labels.rs
```diff
@@ -0,0 +1,260 @@
...
+                // Merge assigned labels from the two datasets
+                let assigned = merge_assigned_labels(
+                    target_assigned,
+                    source_assigned,
+                    &transferred_label_parent,
+                );
+                let new_labelling = NewLabelling {
+                    assigned,
+                    dismissed,
+                };
```
davidhewitt, line 123: Couldn't help myself again :golf: 

```suggestion
                let target_labelling = target_labellings_to_update
                    .remove(&uid)
                    .expect("Tried to make a new labelling for a comment we don't need to update.");

                // Merge assigned labels from the two datasets
                let assigned = merge_assigned_labels(
                    target_labelling.assigned,
                    source_assigned,
                    &transferred_label_parent,
                );
                let new_labelling = NewLabelling {
                    assigned,
                    target_labelling.dismissed,
                };
```

cli/src/commands/alpha/take_more_labels.rs
```diff
@@ -0,0 +1,260 @@
...
+                        &target_dataset.full_name(),
+                        &uid,
+                        Some(&new_labelling),
+                        None,
+                    )
+                    .chain_err(|| ErrorKind::Client("Updating labelling failed.".into()))?;
+                statistics.add_labellings_updated(1);
+            }
+
+            let _progress = make_progress()?;
```
davidhewitt, line 137: Maybe there should just be a single top-level progress bar across all source ids?

cli/src/commands/alpha/take_more_labels.rs
```diff
@@ -0,0 +1,260 @@
...
+        .chain_err(|| ErrorKind::Client("Getting source dataset failed.".into()))?;
+    let target_dataset = client
+        .get_dataset(target_dataset.clone())
+        .chain_err(|| ErrorKind::Client("Getting target dataset failed.".into()))?;
+
+    let statistics = Arc::new(Statistics::default());
+    let make_progress = || -> Result<Progress> { Ok(get_take_more_labels_progress(&statistics)) };
+    make_progress()?;
+
+    for source_id in target_dataset.source_ids.iter() {
```
davidhewitt, line 64: Does this operation potentially only need to be performed for sources which are included in both datasets?

davidhewitt (commented):
This seems a lot like a copy to me, filtered down to a target label. Should we just call it "copy-labels"? 

Being new to this feature I've probably added a lot of stupid questions. Also done a bit of code golfing where it looked like let-destructuring led to more verbose code than needed.

davidhewitt:
> The one question I have around this is how to manage testing/checking these features in CI. Obviously I can duplicate a bunch of cargo clippy and cargo test features in the check script to add the --all-features flag, but is there a better way you know of?

Sorry, originally missed this off. How about just adding `--all-features` to the CI scripts rather than having separate invocations with and without?

cli/Cargo.toml
```diff
@@ -30,3 +30,12 @@ serde_json = "1.0.55"
 structopt = { version = "0.3.15", default-features = false }
 
 reinfer-client = { version = "0.4.0", path = "../api" }
+
+[features]
+default = []
+
+# Enable experimental features that may break at any time.
+experimental = ["alpha"]
```
mcobzarenco, line 38: The `experimental` feature is not used for anything, is it?

mcobzarenco (approved):
Thanks for changing to using conditional compilation :+1: 

Deferring to @davidhewitt on actual code review :smile:

cli/Cargo.toml
davidhewitt: I guess it's nice as a meta-feature so that you can just write `--features experimental` instead of `--features "alpha beta"` etc.

cli/src/commands/create/comments.rs
```diff
@@ -297,6 +297,7 @@ fn upload_batch(
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
tommilligan: Do you mean 'why wasn't this caught already?' I get a failure from running `./scripts/check` locally. Agreed if this was introduced in an earlier commit we should check out CI, but I can't see anything obvious in the setup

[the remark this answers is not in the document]
davidhewitt: See #25

tommilligan:
Closed as obsolete, can be resurrected if anyone is interested.

PR #23, opened 2020-08-25
task: improve invalid source name error message

davidhewitt:
This improves the output of `re create source source-missing-owner` from:

```
E Operation to create a source has failed
E  |- API request failed with 405 Method Not Allowed: 405 Method Not Allowed
```

to:

```
error: Invalid value for '<source-name>': Expected <owner>/<name> or a source id, got: no-owner
```

The best way to test this was via an integration test, so please merge #22 first. After then I'll rebase this PR to isolate just the relevant changes (from `89aea98`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # next
 
+## Changed
+- `create source`: Improve error message when specifying an invalid source name.
+
 # v0.4.0
 
 ## Added
--- a/api/src/resources/source.rs
+++ b/api/src/resources/source.rs
@@ -36,6 +36,21 @@ pub struct Name(pub String);
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct FullName(pub String);
 
+impl FromStr for FullName {
+    type Err = Error;
+
+    fn from_str(string: &str) -> Result<Self> {
+        if string.split('/').count() == 2 {
+            Ok(FullName(string.into()))
+        } else {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+# next
+
+## Added
+- `get current-user`: Add `-s` flag to print just the username and no other output.
+
+## Changed
+- `create source`: Improve error message when specifying an invalid source name.
+
+
```
davidhewitt, line 9: I probably added one too many newlines here.

cli/src/commands/get.rs
```diff
@@ -353,14 +357,19 @@ pub fn run(get_args: &GetArgs, client: Client) -> Result<()> {
...
             let user = client.get_current_user().chain_err(|| {
                 ErrorKind::Client("Operation to get the current user has failed.".into())
             })?;
-            match output {
-                OutputFormat::Table => print_users_table(&[user]),
-                OutputFormat::Json => {
-                    print_resources_as_json(iter::once(&user), io::stdout().lock())?
+
+            if *short {
+                println!("{}", user.username.0);
```
mcobzarenco, line 366: I think `println!` panics if the write fails -- using `io:stdout()` + `write()` would return an error (when piping output this not a theoretical problem)

mcobzarenco (approved):
This looks great, really :100: for adding integration tests -- I've reviewed both commits here just one nit re: `println!`

PR #24, opened 2020-08-28
task: Add retry on timeout for multi-request cmds

davidhewitt:
Add timeout handling at the `reinfer_client` level for API requests. It's disabled by default in the library config.

The CLI enables it to what I call "automatic" timeout handling, which attempts to retry all requests except for the very first. This should mean that if the first request fails with a timeout, the CLI assumes the configured endpoint is invalid and won't retry, leading to a nice quick failure.

For longer-running operations spanning multiple requests, this should lead to nice stability in the face of network connection blips.

This is what stderr looks like when the retry kicks in:

```
W error sending request for url (https://foobarwat.com/qux): operation timed out - retrying after 10s.
W error sending request for url (https://foobarwat.com/qux): operation timed out - retrying after 20s.
W error sending request for url (https://foobarwat.com/qux): operation timed out - retrying after 40s.
E Operation to get the current user has failed.
E  |- An unknown error has occurred: GET operation failed.
E  |- error sending request for url (https://foobarwat.com/qux): operation timed out
```

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -1,18 +1,20 @@
 #![deny(clippy::all)]
 pub mod errors;
 pub mod resources;
+pub mod retry;
 
 use chrono::{DateTime, Utc};
 use failchain::ResultExt;
 use lazy_static::lazy_static;
 use log::debug;
 use reqwest::{
-    blocking::Client as HttpClient,
+    blocking::{Client as HttpClient, Response as HttpResponse},
     header::{HeaderMap, HeaderName, HeaderValue},
-    IntoUrl, Url,
+    IntoUrl, Result as ReqwestResult, Url,
 };
 use serde::{Deserialize, Serialize};
 use serde_json::json;
+use std::cell::Cell;
 use std::fmt::Display;
 
 use crate::resources::{
@@ -45,6 +47,8 @@ use crate::resources::{
     },
     ApiError, EmptySuccess, Response, SimpleApiError,
 };
+
+use crate::retry::{Retrier, RetryConfig};
 
 pub use crate::{
     errors::{Error, ErrorKind, Result},
@@ -87,6 +91,9 @@ pub struct Config {
     pub endpoint: Url,
     pub token: Token,
     pub accept_invalid_certificates: bool,
+    /// Retry settings to use, if any. This will apply to all requests except for POST requests
+    /// which are not idempotent (as they cannot be naively retried).
+    pub retry_config: Option<RetryConfig>,
 }
 
 impl Default for Config {
@@ -95,6 +102,7 @@ impl Default for Config {
             endpoint: DEFAULT_ENDPOINT.clone(),
             token: Token("".to_owned()),
             accept_invalid_certificates: false,
+            retry_config: None,
         }
     }
 }
@@ -104,6 +112,7 @@ pub struct Client {
     endpoints: Endpoints,
     http_client: HttpClient,
     headers: HeaderMap,
+    retrier: Option<Retrier>,
 }
 
 #[derive(Serialize)]
@@ -135,10 +144,12 @@ impl Client {
         let http_client = build_http_client(&config)?;
         let headers = build_headers(&config)?;
         let endpoints = Endpoints::new(config.endpoint)?;
+        let retrier = config.retry_config.map(Retrier::new);
         Ok(Client {
             endpoints,
             http_client,
             headers,
+            retrier,
         })
     }
 
@@ -252,6 +263,7 @@ impl Client {
         Ok(self.post::<_, _, _, SimpleApiError>(
             self.endpoints.sync_comments(source_name)?,
             SyncCommentsRequest { comments },
+            Retry::Yes,
         )?)
     }
 
@@ -371,6 +383,7 @@ impl Client {
                 labelling,
                 entities,
             },
+            Retry::No,
         )?)
     }
 
@@ -394,6 +407,7 @@ impl Client {
                 filter,
                 continuation,
             },
+            Retry::No,
         )?)
     }
 
@@ -423,6 +437,7 @@ impl Client {
             .post::<_, _, GetStatisticsResponse, SimpleApiError>(
                 self.endpoints.statistics(dataset_name)?,
                 json!({}),
+                Retry::No,
             )?
             .statistics)
     }
@@ -486,6 +501,7 @@ impl Client {
         self.post::<_, _, _, SimpleApiError>(
             self.endpoints.trigger_fetch(trigger_name)?,
             TriggerFetchRequest { size },
+            Retry::No,
         )
     }
 
@@ -497,6 +513,7 @@ impl Client {
         self.post::<_, _, serde::de::IgnoredAny, SimpleApiError>(
             self.endpoints.trigger_advance(trigger_name)?,
             TriggerAdvanceRequest { sequence_id },
+            Retry::No,
         )?;
         Ok(())
     }
@@ -511,22 +528,25 @@ impl Client {
             TriggerResetRequest {
                 to_comment_created_at,
             },
+            Retry::No,
         )?;
         Ok(())
     }
 
     fn get<LocationT, SuccessT, ErrorT>(&self, url: LocationT) -> Result<SuccessT>
     where
-        LocationT: IntoUrl + Display,
+        LocationT: IntoUrl + Display + Clone,
         for<'de> SuccessT: Deserialize<'de>,
         for<'de> ErrorT: Deserialize<'de> + ApiError,
     {
         debug!("Attempting GET `{}`", url);
         let http_response = self
-            .http_client
-            .get(url)
-            .headers(self.headers.clone())
-            .send()
+            .with_retries(|| {
+                self.http_client
+                    .get(url.clone())
+                    .headers(self.headers.clone())
+                    .send()
+            })
             .chain_err(|| ErrorKind::Unknown {
                 message: "GET operation failed.".to_owned(),
             })?;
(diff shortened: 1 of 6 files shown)


cli/src/main.rs
```diff
@@ -90,10 +93,20 @@ fn client_from_args(args: &Args, config: &ReinferConfig) -> Result<Client> {
...
     }
 
+    // Retry everything but the very first request.
+    // Retry wait schedule is [10s, 20s, 40s, fail]. (Plus the time for each attempt to timeout.)
+    let timeout_handling = TimeoutHandling {
+        retry_on_timeout: RetryOnTimeout::Automatic,
+        max_retry_count: 3,
+        base_wait: std::time::Duration::from_secs_f64(10.0),
+        backoff_factor: 2.0,
+    };
```
davidhewitt, line 103: Willing to accept bikeshedding on whether these defaults are good, and also whether they should be configurable via global cmdline arguments (or environment variables).

api/src/lib.rs
```diff
@@ -567,15 +579,17 @@ impl Client {
...
-            .http_client
-            .delete(url)
-            .headers(self.headers.clone())
-            .send()
+            .handle_timeouts(|| {
+                self.http_client
+                    .delete(url.clone())
+                    .headers(self.headers.clone())
+                    .send()
+            })
```
cristicbz, line 592: This will fail with a 404 if the first timed out request had actually succeded (and the timeout was on the response side). Not sure how to handle non-idempotent requests---we could just not wrap them or explicitly succeed follow up requests on 404 for deletion (this would have to make the `handle_timeouts` a leakier abstraction)

api/src/lib.rs
```diff
@@ -593,18 +607,20 @@ impl Client {
...
-            .headers(self.headers.clone())
-            .json(&request)
-            .send()
+            .handle_timeouts(|| {
+                self.http_client
+                    .post(url.clone())
+                    .headers(self.headers.clone())
+                    .json(&request)
+                    .send()
+            })
```
cristicbz, line 623: Our requests are generally idempotent, but a blindly retrying a `POST` request seems a bit dangerous---should we make this a parameter to `post` so that each route can decide for itself whether it's safe to retry?

cli/src/main.rs
cristicbz: There's a lot of bikeshedding to do on backoff retries, but I'm happy to have this in as is. Maybe I'd default to 5.0s rather than 10, but whatever.

cristicbz (commented):
I left a handful of comments, I'm mostly concerned about idempotency. One other question is whether we should not also retry on 5xx errors? I think one of the big things from @margold was failures due to nginx 504 (Gateway Timeout).

davidhewitt:
Thanks for the comments. I think it makes sense to also retry on 5xx; I've done this kind of thing in the past for both timeouts and 5xxs. I think the same design of only retry the 2nd request and later is still suitable.

Will re-work this :+1:

davidhewitt:
Rebased on current master, and reimplemented the vast majority of this to also apply to 5xx errors. Taken on the points w.r.t delete and post.

Ready for another look.

api/src/lib.rs
```diff
@@ -584,30 +610,49 @@ impl Client {
             .json::<Response<EmptySuccess, ErrorT>>()
             .chain_err(|| ErrorKind::BadJsonResponse)?
             .into_result(status)
-            .map(|_| ())
+            .map_or_else(
+                // Ignore 404 not found if the request had to be re-tried - assume the target
+                // object was deleted on a previous incomplete request.
+                |e| {
```
cristicbz, line 616: nit. As a matter of style we stay away from single-letter variables. It's easy for things to get out of hand pretty quickly even if the original code is pretty straight-forward.

So here for instance, `error` would be preferred.

api/src/lib.rs
```diff
@@ -584,30 +610,49 @@ impl Client {
...
+                    if attempts.get() > 1 && status == 404 {
```
cristicbz, line 617: nit. you should use `reqwest::StatusCode::NOT_FOUND`

api/src/lib.rs
```diff
@@ -642,6 +689,16 @@ impl Client {
             .chain_err(|| ErrorKind::BadJsonResponse)?
             .into_result(status)
     }
+
+    fn with_retries<F>(&self, f: F) -> ReqwestResult<HttpResponse>
```
cristicbz, line 693: same nit on `f`, maybe `send_request` instead? I'm slightly less fussed about `F`, but even there we tend to do `SendRequestT`

api/src/retry.rs
```diff
@@ -0,0 +1,204 @@
...
+impl Retrier {
+    pub fn new(config: RetryConfig) -> Self {
+        Self {
+            config,
+            is_first_request: AtomicBool::new(true),
+        }
+    }
+
+    pub fn with_retries(&self, f: impl Fn() -> Result<Response>) -> Result<Response> {
+        if self.is_first_request.swap(false, Relaxed)
```
cristicbz, line 46: Why `Relaxed`?  Seems like a pretty insignificant cost to use `SeqCst` here and saves the reader having to reason about the correctness and potentially prevents subtle bugs. Happy to have my mind changed, but my experience with C++ atomics means my stance is that not using `SeqCst` is and optimisation, and therefore has to be justified by performance numbers.

api/src/lib.rs
```diff
@@ -584,30 +610,49 @@ impl Client {
...
+                },
+                |_| Ok(()),
+            )
     }
 
     fn post<LocationT, RequestT, SuccessT, ErrorT>(
         &self,
         url: LocationT,
         request: RequestT,
+        retry: bool,
```
cristicbz, line 631: This is a nit, but my preference would be to have `enum Retry { No, Yes }` and take that as argument, making calls to `post` more readable, `post(endpoint, request, Retry::Yes)` over `post(endpoint, request, true)`.

api/src/retry.rs
```diff
@@ -0,0 +1,204 @@
...
+
+impl Retrier {
+    pub fn new(config: RetryConfig) -> Self {
+        Self {
+            config,
+            is_first_request: AtomicBool::new(true),
+        }
+    }
+
+    pub fn with_retries(&self, f: impl Fn() -> Result<Response>) -> Result<Response> {
```
cristicbz, line 45: same nit on `f`

api/src/retry.rs
```diff
@@ -0,0 +1,204 @@
...
+                ($src:expr) => {{
+                    let wait_factor = self.config.backoff_factor.powi(i.into());
+                    let duration = self.config.base_wait.mul_f64(wait_factor);
+                    log::warn!("{} - retrying after {:?}.", $src, duration);
+                    sleep(duration)
+                }};
+            };
+
+            match f() {
+                Ok(r) if r.status().is_server_error() => {
```
cristicbz, line 63: `r`

api/src/retry.rs
```diff
@@ -0,0 +1,204 @@
...
+                    warn_and_sleep!(format!("{} for {}", r.status(), r.url()))
+                }
+                Err(e) if e.is_timeout() => warn_and_sleep!(e),
```
cristicbz, line 66: `e`

api/src/retry.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    pub fn with_retries(&self, f: impl Fn() -> Result<Response>) -> Result<Response> {
+        if self.is_first_request.swap(false, Relaxed)
+            && self.config.strategy == RetryStrategy::Automatic
+        {
+            return f();
+        }
+
+        for i in 0..self.config.max_retry_count {
```
cristicbz, line 52: nit `i_retry` rather than `i`? this scales much better when nested for loops come into it.

cristicbz (commented):
Right, this looks great! I have a bunch of nits and I hope you don't hate me too much for it.

I appreciate we don't have a written down style guide, but there aren't many rules that we have that aren't already enforced by linters and formatters: the main ones you may take issue with are our opposition to single-letter variables (and non-standard abbreviations more generally) and opposition to boolean arguments.

api/src/retry.rs
davidhewitt: Gone for `retry_count`.

api/src/lib.rs
davidhewitt: Very happy to use enums instead of bools if that's our preferred style. Increased verbosity is why I didn't opt for it initially, though completely agree it can aid interpretation at the call site.

api/src/retry.rs
davidhewitt: My thinking here was that it's not immediately obvious anyway what behaviour is expected if two threads were to launch "the first request" simultaneously. Probably from a user's perspective it's fine for both of them to fail immediately without retry. This is, after all, just trying to make an interface more friendly rather than doing some work that needs to have guaranteed sychronisation.

So the `Relaxed` was because I literally didn't think it mattered in this case. (I considered `Cell` instead of `AtomicBool` but didn't feel I needed to force this type into being `!Sync`.) I completely agree with you that the other 99% of the time `Relaxed` is scary, so happy to use `SeqCst` here just to prevent people getting into bad habits.

davidhewitt:
All points totally reasonable!

Wrt single-variables I can totally be on board - I never do this in Python but tend to be a little more relaxed in Rust thanks to at least having the type of the variable more easily to hand.

Similarly enums instead of bools is definitely nicer at the call site. This is literally the only place I wish that Rust had keyword args, so that I could just write `retry=true` at the call site :smile:

api/src/retry.rs
```diff
@@ -152,12 +152,12 @@ mod tests {
         ok.assert();
 
         // Retries up to N times on timeout for non-first-requests.
-        for i in 0..10 {
+        for retry_count in 0..10 {
```
cristicbz, line 155: I'm all right with `retry_count` here, but in general I do like the `i_retry` convention. `retry_count` is ambiguous: is this the "max number of retries" or the "current retry number". `i_retry` and `num_retries` is what we tend to default to, but I think there's more variability here in the team than with other things.

cristicbz (approved):
Great stuff, ship it!

PR #29, opened 2020-09-16
api: use new query format for comment uid lists

dan-ri:


api/src/lib.rs
```diff
@@ -1081,19 +1081,13 @@ fn build_headers(config: &Config) -> Result<HeaderMap> {
...
-        }
-    }
-    query_uids
+fn comment_uid_list_query<'a>(
+    comment_uids: impl Iterator<Item = &'a CommentUid>,
+) -> Vec<(&'static str, &'a str)> {
+    // Return a list of pairs ("id", "a"), ("id", "b"), ...
+    // The http client will turn this into a query string of
+    // the form "id=a&id=b&..."
+    comment_uids.map(|uid| ("id", uid.0.as_str())).collect()
```
mcobzarenco, line 1090: A small question re: the API format -- there was a separate PR in platform so maybe it was discussed there, but is there a way to avoid repeating `id`? or no because comma is a valid id? I guess query form params get escaped anyway

mcobzarenco (approved):
Looks good

api/src/lib.rs
dan-ri: Yes, ids can contain any printable ASCII character so this seemed like the most straightforward approach.

PR #31, opened 2020-11-04
feature: support for uploading comments with audio

mcobzarenco:
Simple support for uploading audio files associated with comments -- the stats are kind of broken as we don't include the audio file size in the upload, but much more convenient than curl.

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -7,14 +7,13 @@ use chrono::{DateTime, Utc};
 use lazy_static::lazy_static;
 use log::debug;
 use reqwest::{
-    blocking::{Client as HttpClient, Response as HttpResponse},
+    blocking::{multipart::Form, Client as HttpClient, Response as HttpResponse},
     header::{HeaderMap, HeaderName, HeaderValue},
     IntoUrl, Result as ReqwestResult,
 };
 use serde::{Deserialize, Serialize};
 use serde_json::json;
-use std::cell::Cell;
-use std::fmt::Display;
+use std::{cell::Cell, fmt::Display, path::Path};
 use url::Url;
 
 use crate::resources::{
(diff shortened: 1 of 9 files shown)


api/src/lib.rs
```diff
@@ -972,6 +999,20 @@ impl Endpoints {
...
 
+    fn comment_audio(&self, source_id: &SourceId, comment_id: &CommentId) -> Result<Url> {
+        self.base
+            .join(&format!(
+                "/api/_private/sources/id:{}/comments/{}/audio",
+                source_id.0, comment_id.0
+            ))
+            .chain_err(|| ErrorKind::Unknown {
+                message: format!(
+                    "Could not build sync URL for source id `{}` and comment id `{}`.",
```
davidhewitt, line 1010:
```suggestion
                    "Could not build put audio URL for source id `{}` and comment id `{}`.",
```

api/src/resources/comment.rs
```diff
@@ -361,6 +363,8 @@ pub struct NewAnnotatedComment {
     pub labelling: Option<NewLabelling>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub entities: Option<NewEntities>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub audio_path: Option<PathBuf>,
```
davidhewitt, line 367: Stupid question: I assume that this doesn't round-trip? If so, should it go in some clearly denoted subsection like "upload_extras" ?

cli/src/commands/create/comments.rs
```diff
@@ -279,6 +281,19 @@ fn upload_batch(
         unchanged,
     });
 
+    // Upload audio
+    for (comment_id, audio_path) in audio_paths.iter() {
```
davidhewitt, line 285: Nit: as this argument appears after `annotations` in the arguments list, it might be tidy if similarly this block was after the annotations block. Unless there's a logical reason not to do so?

davidhewitt (approved):
Sorry for the delay in reviewing. Code LGTM. I haven't verified it runs locally, happy to do so if you can send me an audio file to test it with.

I have a couple of tiny nits / questions. Also should this go in some documentation, so that users know they can do this - maybe just the CHANGELOG?

Also, as always - should we perhaps add an integration test for this? That could also help serve as "documentation" :smile:

api/src/lib.rs
mcobzarenco: Well spotted -- I fixed it, without the mention of "put" as this is just the url for the audio content -- it can also be get

mcobzarenco:
> I have a couple of tiny nits / questions. Also should this go in some documentation, so that users know they can do this - maybe just the CHANGELOG?

This is a bit of an internal change tbh as audio is a super alpha feature we don't advertise

api/src/resources/comment.rs
mcobzarenco: I thought it's similar to the existing `annotations` key -- e.g. at the root level, we support 3 keys
 - `comment` (the actual comment content)
 - `annotations` (labels and entitites annotations, they're uploaded separately to a different route and scoped under a comment)
 -  now `audio_path` (also uploaded separately and scoped under a comment), path to a file that contains the audio content. It's unsatisfactory to have this external reference, but bar inline the content of an audio file in json, we can't do much about it
 
> Stupid question: I assume that this doesn't round-trip?

mcobzarenco:
> Also, as always - should we perhaps add an integration test for this? That could also help serve as "documentation" smile

Good point, I added a test

mcobzarenco:
Suuuuper belated thanks for the review (I needed it for a one off, so forgot about it after that, ha!) -- I rebased and addressed your comments apart from updating the changelog for the reason I mention in the comment.

PR #32, opened 2020-11-19
api/cli: add 'delete comments' to delete comments by id

tommilligan:
Adds the ability to delete comments by id in a source.

Example usage:

```bash
cargo run -- delete comments --source org1/naughty-comments plain_comment plain_comment_translated
```

Includes factoring out:

- `id_list_query` as a more generic implmentation
- `get` to reuse `get_query` (to follow the pattern of `delete_query`)

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -192,13 +192,29 @@ impl Client {
             .source)
     }
 
-    /// Delete a new source.
-    pub fn delete_source(&self, source: impl Into<SourceIdentifier> + Clone) -> Result<()> {
-        let source_id = match source.clone().into() {
+    /// Delete a source.
+    pub fn delete_source(&self, source: impl Into<SourceIdentifier>) -> Result<()> {
+        let source_id = match source.into() {
             SourceIdentifier::Id(source_id) => source_id,
-            SourceIdentifier::FullName(_) => self.get_source(source)?.id,
+            source @ SourceIdentifier::FullName(_) => self.get_source(source)?.id,
         };
         self.delete::<_, SimpleApiError>(self.endpoints.source_by_id(&source_id)?)
+    }
+
+    /// Delete comments by id in a source.
+    pub fn delete_comments(
+        &self,
+        source: impl Into<SourceIdentifier>,
+        comments: &[CommentId],
+    ) -> Result<()> {
+        let source_full_name = match source.into() {
+            source @ SourceIdentifier::Id(_) => self.get_source(source)?.full_name(),
+            SourceIdentifier::FullName(source_full_name) => source_full_name,
+        };
+        self.delete_query::<_, _, SimpleApiError>(
+            self.endpoints.comments_v1(&source_full_name)?,
+            Some(&id_list_query(comments.iter().map(|uid| &uid.0))),
+        )
     }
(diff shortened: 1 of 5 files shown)


api/src/lib.rs
```diff
@@ -201,6 +201,22 @@ impl Client {
...
     }
 
+    /// Delete comments by id in a source.
+    pub fn delete_comments(
+        &self,
+        source: impl Into<SourceIdentifier> + Clone,
+        comments: &Vec<CommentId>,
+    ) -> Result<()> {
+        let source_full_name = match source.clone().into() {
+            SourceIdentifier::Id(_) => self.get_source(source)?.full_name(),
```
davidhewitt, line 211: Nit: I think can skip the clone, which can allow you to remove it from the `source` argument bounds above:

```suggestion
        let source_full_name = match source.into() {
            source @ SourceIdentifier::Id(_) => self.get_source(source)?.full_name(),
```

api/src/lib.rs
```diff
@@ -539,28 +555,13 @@ impl Client {
...
-            })
-            .chain_err(|| ErrorKind::Unknown {
-                message: "GET operation failed.".to_owned(),
-            })?;
-        let status = http_response.status();
-        http_response
-            .json::<Response<SuccessT, ErrorT>>()
-            .chain_err(|| ErrorKind::BadJsonResponse)?
-            .into_result(status)
+        self.get_query::<LocationT, (), SuccessT, ErrorT>(url, None)
```
davidhewitt, line 558: :+1: nice cleanup!

api/src/lib.rs
```diff
@@ -571,11 +572,14 @@ impl Client {
...
-                    .query(query)
-                    .send()
+                    .headers(self.headers.clone());
+                if let Some(query) = query {
+                    request = request.query(query);
+                }
+                request.send()
             })
             .chain_err(|| ErrorKind::Unknown {
                 message: "GET operation failed.".to_owned(),
```
davidhewitt, line 585: If you're wanting to shoot for maximum code reuse, this same structure is basically mirrored in the first part of `delete_query` barring `http_client.get` vs `http_client.delete` - could potentially use `http_client.request` to parametrise this bit over the method.

But I'm also fine with leaving this as-is :smile:

davidhewitt (approved):
:+1: lgtm.

Couple of nits/suggestions - also it might be interesting to add an integration test for this. Could either modify the existing comments integration test or add a clone of it just below:

https://github.com/reinfer/cli/blob/c0e01d240e040a6e87da8884dbf6eef7529c816d/cli/tests/test_comments.rs#L6

davidhewitt:
Maybe also worth adding a CHANGELOG entry for this?

api/src/lib.rs
tommilligan: That's cool - didn't know about `@` syntax

api/src/lib.rs
tommilligan: I thought about that, but decided to leave as is :sweat_smile: if we need another `*_query` handler I'll definitely factor it out

davidhewitt (approved):
:100: nice!

PR #34, opened 2020-11-20
tests: require integration tests to be explicit about organisation

davidhewitt:
At the moment to be "smart" the integration tests try to default the organisation to be the same as the username.

Almost always a user doesn't have an organisation which matches their username, so this often fails.

Also implicit behaviour is probably not good. These tests now require the env variable to be set explicitly.


tommilligan (approved):
LGTM

PR #36, opened 2020-11-26
buckets: add transform tag to create and get commands

tommilligan:
We've already been able to specify this, it was just missing.

Will soon be required, so we'll need to have it in the CLI to create buckets.

image

cc @Joe-Prosser if you're interested in how the cli interacts with the api

### DIFF
--- a/cli/tests/common.rs
+++ b/cli/tests/common.rs
@@ -50,6 +50,10 @@ impl TestCli {
         self.output(self.command().args(args))
     }
 
+    pub fn run_and_error(&self, args: impl IntoIterator<Item = impl AsRef<OsStr>>) -> String {
+        self.output_error(self.command().args(args))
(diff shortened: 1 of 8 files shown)


cli/tests/common.rs
```diff
@@ -88,4 +92,17 @@ impl TestCli {
 
         String::from_utf8(output.stdout).unwrap()
     }
+
+    pub fn output_error(&self, command: &mut Command) -> String {
+        let output = command.output().unwrap();
+
+        if output.status.success() {
+            panic!(
+                "suceeded running command (expected failure):\n{}",
```
dan-ri, line 101:
```suggestion
                "succeeded running command (expected failure):\n{}",
```

dan-ri (approved):
LG! :100:

PR #39, opened 2020-12-14
create-dataset: add label copying and model family args

AWilcke:
Adds the `copy_labels_from` and `model_family` arguments for the `create/dataset` route.

### DIFF
--- a/cli/src/commands/create/dataset.rs
+++ b/cli/src/commands/create/dataset.rs
@@ -30,6 +30,14 @@ pub struct CreateDatasetArgs {
     #[structopt(short = "e", long = "entity-defs", default_value = "[]")]
     /// Entity defs to create at dataset creation, as json
     entity_defs: EntityDefs,
+
+    #[structopt(long = "model-family")]
+    /// Model family to use for the new dataset
+    model_family: Option<String>,
+
+    /// Dataset ID of the dataset to copy labels from
+    #[structopt(long = "copy-labels-from")]
+    copy_labels_from: Option<String>,
 }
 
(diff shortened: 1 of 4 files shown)


cli/src/commands/create/dataset.rs
```diff
@@ -66,6 +76,12 @@ pub fn create(client: &Client, args: &CreateDatasetArgs) -> Result<()> {
                 description: description.as_ref().map(|description| description.as_str()),
                 has_sentiment: *has_sentiment,
                 entity_kinds: &entity_kinds,
+                model_family: model_family
+                    .as_ref()
+                    .map(|model_family| model_family.as_str()),
+                copy_labels_from: copy_labels_from
+                    .as_ref()
+                    .map(|copy_labels_from| copy_labels_from.as_str()),
```
davidhewitt, line 84: I think you're looking for `Option::as_deref`:

```suggestion
                model_family: model_family.as_deref(),
                copy_labels_from: copy_labels_from.as_deref(),
```

(looks like you copied this pattern from other uses of Option in this file - maybe tidy the others up too if you're feeling generous?)

davidhewitt (approved):
Thanks very much for this, looks good!

If you're feeling up to it, would be great to add some integration tests for the `create dataset` command at the same time. Such tests would invoke the cli binary and create dataset(s) in a test organisation of your choice. See https://github.com/reinfer/cli/blob/master/cli/tests/test_sources.rs for the tests I added for sources.

tommilligan (approved):
Tests LGTM!

tommilligan (approved):

PR #41, opened 2020-12-19
create bucket: validate bucket name

Joe-Prosser:
Validates that a bucket name is well formed on creation.

### DIFF
--- a/api/src/resources/bucket.rs
+++ b/api/src/resources/bucket.rs
@@ -1,4 +1,6 @@
 use chrono::{DateTime, Utc};
+use lazy_static::lazy_static;
+use regex::Regex;
 use serde::{Deserialize, Serialize};
 use std::fmt::{Display, Formatter, Result as FmtResult};
 use std::str::FromStr;
@@ -7,6 +9,11 @@ use crate::{
     errors::{Error, ErrorKind, Result},
     resources::user::Username,
 };
+
+lazy_static! {
+    static ref FULL_NAME_REGEX: Regex =
+        Regex::new("^[A-Za-z0-9-_]{1,256}/[A-Za-z0-9-_]{1,256}$").unwrap();
+}
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Bucket {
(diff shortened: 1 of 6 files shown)


api/src/resources/bucket.rs
```diff
@@ -70,11 +71,12 @@ impl From<Id> for Identifier {
 
 impl FromStr for Identifier {
     type Err = Error;
-
     fn from_str(string: &str) -> Result<Self> {
+        let full_name_regex = Regex::new("^[^/]+/[^/]+$").unwrap();
```
jcalero, line 75: Do we not want to at least attempt to mimic the constraints in the store for this regex? Currently the bucket name regex in the backend is `^[A-Za-z0-9-_]{1,256}/[A-Za-z0-9-_]{1,256}$`. We could simplify it and skip the size and just do `+`, but I think we should at least constraints the character set, no?

api/src/resources/bucket.rs
```diff
@@ -168,3 +170,19 @@ impl Display for BucketType {
         }
     }
 }
+
+impl FromStr for FullName {
+    type Err = Error;
+    fn from_str(string: &str) -> Result<Self> {
+        let full_name_regex = Regex::new("^[^/]+/[^/]+$").unwrap();
```
jcalero, line 177: Same here, although, is there a way we can make the regex a constant so we don't have to modify it in two places? Also makes sure we don't have mismatching regexes.

api/src/resources/bucket.rs
jcalero: PS. I know we don't need to be very precise here since the backend will catch the errors anyway, but since we're here already I don't see why not? I doubt we'll ever relax the backend regex more than it is so we might as well fail in the CLI.

jcalero (commented):
Good stuff! (from my limited rust knowledge)

Just a couple of small suggestions

jcalero (approved):
Looks good to me!

tommilligan:
For https://github.com/reinfer/platform/issues/9629

LGTM, thanks for fixing this!

PR #46, opened 2021-02-05
cli: Add validator check for Dataset name

scallaway:
Adds validation to input for the Dataset name to ensure that the URL that's constructed is what the API expects.

### DIFF
--- a/cli/src/commands/create/dataset.rs
+++ b/cli/src/commands/create/dataset.rs
@@ -1,13 +1,28 @@
 use anyhow::{Context, Result};
+use lazy_static::lazy_static;
 use log::info;
+use regex::Regex;
 use reinfer_client::{
     resources::dataset::EntityDefs, Client, DatasetFullName, NewDataset, SourceIdentifier,
 };
 use structopt::StructOpt;
 
+lazy_static! {
+    static ref RX_FULL_NAME: Regex =
+        Regex::new("^[A-Za-z0-9-_.]{1,1024}/[A-Za-z0-9-_]{1,256}").unwrap();
+}
+
+/// Ensures the name provided conforms to the <owner>/<name> structure
+fn validate_dataset_name(s: String) -> std::result::Result<(), String> {
+    match RX_FULL_NAME.is_match(&s) {
+        true => Ok(()),
+        false => Err(s),
+    }
+}
+
 #[derive(Debug, StructOpt)]
 pub struct CreateDatasetArgs {
-    #[structopt(name = "dataset-name")]
+    #[structopt(name = "owner-name/dataset-name", validator = validate_dataset_name)]
     /// Full name of the new dataset <owner>/<name>
     name: String,
 
--- a/cli/src/commands/create/dataset.rs
+++ b/cli/src/commands/create/dataset.rs
@@ -1,23 +1,22 @@
 use anyhow::{Context, Result};
-use lazy_static::lazy_static;
 use log::info;
+use once_cell::sync::Lazy;
 use regex::Regex;
 use reinfer_client::{
     resources::dataset::EntityDefs, Client, DatasetFullName, NewDataset, SourceIdentifier,
 };
 use structopt::StructOpt;
 
-lazy_static! {
-    static ref RX_FULL_NAME: Regex =
-        Regex::new("^[A-Za-z0-9-_.]{1,1024}/[A-Za-z0-9-_]{1,256}").unwrap();
-}
+static RX_FULL_NAME: Lazy<Regex> =
+    Lazy::new(|| Regex::new("^[A-Za-z0-9-_.]{1,1024}/[A-Za-z0-9-_]{1,256}").unwrap());
 
 /// Ensures the name provided conforms to the <owner>/<name> structure
-fn validate_dataset_name(s: String) -> std::result::Result<(), String> {
-    match RX_FULL_NAME.is_match(&s) {
-        true => Ok(()),
-        false => Err(s),
+fn validate_dataset_name(s: String) -> Result<(), String> {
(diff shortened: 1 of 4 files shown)


cli/src/commands/create/dataset.rs
```diff
@@ -1,15 +1,30 @@
...
+    static ref RX_FULL_NAME: Regex =
+        Regex::new("^[A-Za-z0-9-_.]{1,1024}/[A-Za-z0-9-_]{1,256}").unwrap();
+}
+
+/// Ensures the name provided conforms to the <owner>/<name> structure
+fn validate_dataset_name(s: String) -> std::result::Result<(), String> {
+    match RX_FULL_NAME.is_match(&s) {
+        true => Ok(()),
+        false => Err(s),
+    }
```
davidhewitt, line 22: I personally prefer `if x { foo } else { bar }` rather than matching on boolean.

Also possible is to use something along the lines of `failchain::ensure!(RX_FULL_NAME.is_match(&s), ...)` as you're returning an error on the bad case - but that's probably not very ergonomic until we migrate from `failchain` to `anyhow`.

cli/src/commands/create/dataset.rs
```diff
@@ -1,15 +1,30 @@
...
     resources::comment::EntityKind, Client, DatasetFullName, NewDataset, SourceIdentifier,
 };
 use structopt::StructOpt;
 
 use crate::errors::{ErrorKind, Result};
 
+lazy_static! {
+    static ref RX_FULL_NAME: Regex =
+        Regex::new("^[A-Za-z0-9-_.]{1,1024}/[A-Za-z0-9-_]{1,256}").unwrap();
+}
```
davidhewitt, line 15: Optional: we're tending towards using `once_cell` instead of `lazy_static`: no need for macro magic and also `once_cell` is potentially on the way to the standard library.

Could use `once_cell::sync::Lazy` if you fancy trying it out. (Or indeed migrate the whole cli project to `once_cell` in a separate PR?)

davidhewitt (approved):
Thanks, a nice usability improvement! Optional if you think this is worth mentioning in the changelog.

Also two super nitty suggestions below.

davidhewitt:
Also, if you're up for a little extra challenge: add a test in `cli/tests/` for this. There's a number of tests in that directory which execute the cli binary and verify its output. Here you could run with invalid name and verify it gives the helpful error message. (Could even add a test to successfully create a dataset, if you're up for it!)

cli/src/commands/create/dataset.rs
```diff
@@ -1,15 +1,30 @@
...
 use reinfer_client::{
     resources::comment::EntityKind, Client, DatasetFullName, NewDataset, SourceIdentifier,
 };
 use structopt::StructOpt;
 
 use crate::errors::{ErrorKind, Result};
 
+lazy_static! {
+    static ref RX_FULL_NAME: Regex =
+        Regex::new("^[A-Za-z0-9-_.]{1,1024}/[A-Za-z0-9-_]{1,256}").unwrap();
```
tommilligan, line 14: I thought we would want to limit this to a shorter name overall due to API constraints? cc: @annesylvie and her recent api validation work.

I'm also happy to leave this more permissible and just return the API error if the name is too long.

tommilligan: Currently, with this regex you can get an error like:
```bash
$ cargo run -- create dataset org1/some-incredibly-long-dataset-name-that-will-cause-a-500-error-in-the-api-because-its-over-127-chars-long-no-really-now-its-over-the-limit
E Operation to create a dataset has failed.
E  |- API request failed with 422 Unprocessable Entity: Dataset name is too long; the maximum length of the string org1/<dataset_name> is of 127 characters
```
which I think is absolutely fine

tommilligan (approved):

cli/src/commands/create/dataset.rs
scallaway: Maybe I combined the wrong regex? I pulled this from the API code, happy to update it to something even more restrictive if that's the way to go.

annesylvie: I did indeed add a new constraint on the length of the string `<org>/<dataset_name>` that is accepted by the api; I guess this was merged after you started working on this!

The regex itself is unchanged; we check for the length of the string separately. Up to you if you also want to implement this behaviour @scallaway. It can also be another PR if you prefer.

tommilligan: I think in that case it's fine to leave the regex as is. We expect to surface API errors directly to users in a lot of cases, the validation in the CLI should be expected to be light (similar to frontend validation)

cli/src/commands/create/dataset.rs
scallaway: Yep, makes sense to just use an `if` here rather than a `match`.

I don't think that we can use `anyhow` here since the expected return type of a validator function for `structopt` is `Result<(), String>` and so a conversion to an `anyhow` Error doesn't work here.

cli/src/commands/create/dataset.rs
scallaway: I think converting this here to use `once_cell` rather than `lazy_static` is a good idea, and I'll do it in this PR.

Should probably leave conversion of the rest of the project to `once_cell` to a different PR like you say!

scallaway:
@tommilligan @davidhewitt 

Re-requesting reviews from both of you as I've added a new file `test_datasets` since that didn't exist previously, and modeled it based off `test_sources`. Plus I've addressed the suggestions from David!

tommilligan (approved):
LGTM, thanks for doing the tests

davidhewitt (approved):
Looks great. Thanks very much for improving this!

PR #48, opened 2021-02-16
Allow users to specify EntityDef-s at dataset creation

annesylvie:
Currently blocked on deploying the `api` change allowing to use the builtin names inside `inherits_from`.

We currently support the `-e` flag at dataset creation to enable specific entity kinds. However, as we are now in a world that uses `EntityDef`-s, the cli needs to be able to create these more complex objects.

One thing I had to decide was how the user would be passing these objects to the api. I opted for providing the corresponding json object, à la `curl`. Happy to go another way if someone has a better suggestion.

Example use:
```
re create dataset org/example-dataset -s org/example-source --has-sentiment false -e '{"name":"trainable_org","title":"Custom Organisation","inherits_from":["org"],"trainable":true}' -e '{"name":"non_trainable_person","title":"Basic Person","inherits_from":["person"],"trainable":false}'
```
One `-e` flag needs to be passed for each entity def.

### DIFF
--- a/api/src/error.rs
+++ b/api/src/error.rs
@@ -43,6 +43,17 @@ pub enum Error {
     #[error("Expected a valid bucket type, got: {}", bucket_type)]
     BadBucketType { bucket_type: String },
 
+    #[error(
+        "Expected valid json for entity def. Got: '{}', which failed because: '{}'",
+        entity_def,
+        source
+    )]
+    BadEntityDef {
+        entity_def: String,
+        #[source]
+        source: ::serde_json::Error,
+    },
+
     #[error("Could not parse JSON response.")]
     BadJsonResponse(#[source] reqwest::Error),
 
--- a/api/src/resources/comment.rs
+++ b/api/src/resources/comment.rs
@@ -470,7 +470,7 @@ fn should_skip_serializing_entities(maybe_entities: &Option<Entities>) -> bool {
 
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
 pub struct NewEntity {
-    pub kind: EntityKind,
+    pub name: EntityName,
     pub formatted_value: String,
     pub span: NewEntitySpan,
 }
@@ -485,7 +485,7 @@ pub struct NewEntitySpan {
 
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
 pub struct Entity {
-    pub kind: EntityKind,
+    pub name: EntityName,
     pub formatted_value: String,
     pub span: EntitySpan,
 }
@@ -501,15 +501,7 @@ pub struct EntitySpan {
 }
 
 #[derive(Debug, PartialEq, Clone, Deserialize, Serialize)]
-pub struct EntityKind(pub String);
-
-impl FromStr for EntityKind {
-    type Err = Error;
-
-    fn from_str(string: &str) -> Result<Self> {
-        Ok(EntityKind(string.to_owned()))
-    }
-}
+pub struct EntityName(pub String);
 
 #[inline]
 fn strip_prefix(string: &mut String, prefix: &str) -> bool {
--- a/cli/src/commands/create/dataset.rs
+++ b/cli/src/commands/create/dataset.rs
@@ -1,7 +1,7 @@
 use anyhow::{Context, Result};
 use log::info;
 use reinfer_client::{
-    resources::comment::EntityKind, Client, DatasetFullName, NewDataset, SourceIdentifier,
+    resources::dataset::EntityDefs, Client, DatasetFullName, NewDataset, SourceIdentifier,
 };
 use structopt::StructOpt;
 
@@ -27,9 +27,9 @@ pub struct CreateDatasetArgs {
     /// Names or ids of the sources in the dataset
     sources: Vec<SourceIdentifier>,
 
-    #[structopt(short = "e", long = "entity-kinds")]
-    /// Entity kinds to enable at dataset creation (pass one flag for each kind)
-    entity_kinds: Vec<EntityKind>,
+    #[structopt(short = "e", long = "entity-defs", default_value = "[]")]
+    /// Entity defs to create at dataset creation, as json
+    entity_defs: EntityDefs,
 }
 
 pub fn create(client: &Client, args: &CreateDatasetArgs) -> Result<()> {
@@ -39,7 +39,7 @@ pub fn create(client: &Client, args: &CreateDatasetArgs) -> Result<()> {
         ref description,
         ref has_sentiment,
         ref sources,
-        ref entity_kinds,
+        ref entity_defs,
     } = *args;
 
     let source_ids = {
@@ -63,7 +63,7 @@ pub fn create(client: &Client, args: &CreateDatasetArgs) -> Result<()> {
                 title: title.as_ref().map(|title| title.as_str()),
                 description: description.as_ref().map(|description| description.as_str()),
                 has_sentiment: *has_sentiment,
-                entity_kinds: &entity_kinds,
+                entity_defs,
             },
         )
         .context("Operation to create a dataset has failed.")?;
--- a/api/src/resources/dataset.rs
+++ b/api/src/resources/dataset.rs
@@ -4,8 +4,9 @@ use std::str::FromStr;
 
 use crate::{
     error::{Error, Result},
-    resources::{comment::EntityKind, source::Id as SourceId, user::Username},
+    resources::{comment::EntityName, source::Id as SourceId, user::Username},
 };
+use serde_json::Error as JsonError;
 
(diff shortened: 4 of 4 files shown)


api/src/resources/comment.rs
```diff
@@ -5,7 +5,7 @@ use serde::{
     ser::{SerializeMap, Serializer},
     Deserialize, Serialize,
 };
-use serde_json::{json, Value as JsonValue};
+use serde_json::{json, Error as SerdeJsonError, Value as JsonValue};
```
davidhewitt, line 8: I might be tempted to just call this `JsonError` (to match `JsonValue`).

cli/src/commands/create/dataset.rs
```diff
@@ -27,9 +27,9 @@ pub struct CreateDatasetArgs {
     /// Names or ids of the sources in the dataset
     sources: Vec<SourceIdentifier>,
 
-    #[structopt(short = "e", long = "entity-kinds")]
-    /// Entity kinds to enable at dataset creation (pass one flag for each kind)
-    entity_kinds: Vec<EntityKind>,
+    #[structopt(short = "e", long = "entity-def")]
```
davidhewitt, line 30: Does it accept multiple values space-separated?

api/src/resources/comment.rs
```diff
@@ -501,13 +501,23 @@ pub struct EntitySpan {
...
+}
+
+impl FromStr for EntityDef {
     type Err = Error;
 
     fn from_str(string: &str) -> Result<Self> {
-        Ok(EntityKind(string.to_owned()))
+        serde_json::from_str(string).map_err(|_: SerdeJsonError| Error::BadEntityDefs {
+            entity_defs: string.to_string(),
+        })
```
tommilligan, line 520: Couple of things here:

- if you're not using the error, you don't need to type it; you can lose the `SerdeJsonError` annotation and corresponding import (I actually don't think you need the type even if you're using it here, Rust should infer it)
- However, I would use the error to show more details to the user!

```suggestion
        serde_json::from_str(string).map_err(|error: SerdeJsonError| Error::BadEntityDef {
            entity_def: string.to_owned(),
            source: error,
        })
```
and see next comment for Error changes

tommilligan (changes requested):
Nice work customising the parsing, that's a really cool idea. I think the error handling can be clearer though (we should still expose the underlying JSON parsing error to the user).

cli/src/commands/create/dataset.rs
annesylvie: Nope, it does not. I did try to add a struct called `EntityDefs` which was just made of a single `Vec<EntityDef>` but then the user would have to provide a json of the form
```
{
    "entity_defs": { ... }
}
```
which I found cumbersome, and I'd need to remove this layer before passing to the api, so I decided to go for the option of providing one flag per entity def. Happy to pair to see how to make it work though!

api/src/error.rs
```diff
@@ -43,6 +43,9 @@ pub enum Error {
     #[error("Expected a valid bucket type, got: {}", bucket_type)]
     BadBucketType { bucket_type: String },
 
+    #[error("Expected valid json for entity defs, got: {}", entity_defs)]
+    BadEntityDefs { entity_defs: String },
```
tommilligan, line 47: Github lost my comment...

```suggestion
    #[error(
        "Expected valid json for entity def. Got: '{}', which failed because: '{}'",
        entity_def,
        source
    )]
    BadEntityDef {
        entity_def: String,
        #[source]
        source: ::serde_json::Error,
    },
```

- nit: rename as it's just an error for a single entity def
- capture and display the underlying error message. Good for traceability, and more info to the user

```bash
cargo run -- create dataset org1/bar -e '{"foo":"bar"}'
error: Invalid value for '--entity-def <entity-defs>...': Expected valid json for entity def. Got: '{"foo":"bar"}', which failed because: 'missing field `name` at line 1 column 13'
```

api/src/resources/comment.rs
annesylvie: I did follow what I could see the convention was for error handling (which I thought was a bit strange to be honest). I see in our errors that we do not resurface the error but instead send back what the user has given as an input.

I do totally agree that surfacing the `json` error does seem like a good idea, so I'll go for your suggestion. :+1:

api/src/error.rs
annesylvie: Excellent! Thanks!

annesylvie: What does `#[source]` do actually?

cli/src/commands/create/dataset.rs
davidhewitt: You probably needed this structure:

```rust
#[derive(Deserialize)]
struct EntityDefs(Vec<EntityDef>);
```

... which I think should deserialize exactly as `[entity_def_1, entity_def_2, ...]` without the wrapping object.

api/src/error.rs
tommilligan: It a) does some magic allowing errors to be nested in this way IIRC, but most importantly it marks this error has having caused the parent one.

This allows methods like `source()` which returns the underlying error. `anyhow` uses this to do things like [`chain`](https://docs.rs/anyhow/1.0.38/anyhow/struct.Error.html#method.chain) which returns an iterator of all the errors that caused this one (kind of like a traceback)

api/src/resources/comment.rs
tommilligan: It's a good point. Showing one error inside another is definitively **not** best practice **in a library**.

Ideally, in your library, you're not really concerned with messages that your user would see. At the point it becomes user visible (i.e. in your **binary**), you would handle this case by doing something like:

```rust
let value = parse_my_cli_args().map_err(|error| {
    match error {
        BadEntityDef { entity_def, source } => anyhow!("Bad entity def {}, caused by {}", entity_def, source),
        _ => error,
    }
});
```

However, the point at which we return a result is inside the `structopt` library, so we can't do that here.

You could argue that this whole parsing-from-string is a `cli` application issue, so the code for this should live there. And then convert whatever you parse into the `api` `EntityDef` struct later. But that is very overly complex for this case!

tommilligan (approved):

cli/src/commands/create/dataset.rs
```diff
@@ -27,9 +27,9 @@ pub struct CreateDatasetArgs {
     /// Names or ids of the sources in the dataset
     sources: Vec<SourceIdentifier>,
 
-    #[structopt(short = "e", long = "entity-kinds")]
-    /// Entity kinds to enable at dataset creation (pass one flag for each kind)
-    entity_kinds: Vec<EntityKind>,
+    #[structopt(short = "e", long = "entity-def", default_value = "[]")]
```
davidhewitt, line 30:
```suggestion
    #[structopt(short = "e", long = "entity-defs", default_value = "[]")]
```

cli/src/commands/create/dataset.rs
```diff
@@ -27,9 +27,9 @@ pub struct CreateDatasetArgs {
...
+    /// Entity def to create at dataset creation, as json. Pass one flag per entity def.
```
davidhewitt, line 31: Outdated comment / help text.

api/src/resources/comment.rs
```diff
@@ -501,13 +501,33 @@ pub struct EntitySpan {
...
     type Err = Error;
 
     fn from_str(string: &str) -> Result<Self> {
-        Ok(EntityKind(string.to_owned()))
+        serde_json::from_str(string).map_err(|error: JsonError| Error::BadEntityDef {
+            entity_def: string.to_owned(),
+            source: error,
+        })
     }
 }
```
davidhewitt, line 532: `EntityDef` and `EntityDefs` seem like they're probably more a concern for `dataset.rs` than `comment.rs`?

davidhewitt (approved):
Just some tiny nits, then lgtm!

cli/src/commands/create/dataset.rs
```diff
@@ -63,7 +63,7 @@ pub fn create(client: &Client, args: &CreateDatasetArgs) -> Result<()> {
                 title: title.as_ref().map(|title| title.as_str()),
                 description: description.as_ref().map(|description| description.as_str()),
                 has_sentiment: *has_sentiment,
-                entity_kinds: &entity_kinds,
+                entity_defs: entity_defs,
```
davidhewitt, line 66: Clippy failing here with lint to use shorthand:

```suggestion
                entity_defs,
```

api/src/resources/comment.rs
annesylvie: Good point, I moved it :+1:

cli/src/commands/create/dataset.rs
annesylvie: It worked, thanks!

PR #49, opened 2021-02-17
tests: panic if any test env var unset

tommilligan:
Found while QAing #39 

The integration test still ran with my default context, even thought the matching environment variable was not set.


AWilcke (approved):
LGTM! No more hitting prod accidentally :see_no_evil:

PR #50, opened 2021-02-17
tests: require test environment to be set

tommilligan:
Replaces #49 as we only need one of the context, or the endpoint + token to be set.


AWilcke (approved):
Nice!

PR #54, opened 2021-02-26
ci: publish binaries for mac

margold:
Tested here: https://github.com/reinfer/cli/runs/2015902856?check_suite_focus=true

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,10 +10,14 @@ env:
 
 jobs:
   publish:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-10.15]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/cache@v2
+        if: ${{ matrix.os != 'macos-10.15' }}
         with:
           path: |
             ~/.cargo/registry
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -10,10 +10,14 @@ env:
 
 jobs:
   publish:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }} 
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-latest]
```
tommilligan, line 16: Let's specify `macos-10.15` rather than `macos-latest` here - [reference](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on)

.github/workflows/publish.yml
```diff
@@ -10,10 +10,14 @@ env:
...
     steps:
       - uses: actions/checkout@v2
       - uses: actions/cache@v2
+        if: ${{ matrix.os != 'macos-latest' }}
```
tommilligan, line 20: I take it this is because caching breaks something on the mac runner?

cli/publish-binaries
```diff
@@ -7,20 +7,24 @@ cd "$(dirname "$0")"
...
+    build_and_upload "x86_64-unknown-linux-gnu"
+
+    # Build and upload deb package
+    cargo deb
+    gsutil cp "../target/debian/reinfer-cli_${CLI_VERSION}_amd64.deb" gs://reinfer-public/cli/debian/
+    ;;
+  "macos-latest")
+    build_and_upload "x86_64-apple-darwin"
+    ;;
+esac
```
tommilligan, line 30: I think you should have a default clause that errors here, something like
```bash
      *)
        >&2 echo "fatal: unknown BUILD_PLATFORM '$BUILD_PLATFORM'"
        exit 1
        ;;
```

scripts/publish-install
```diff
@@ -2,8 +2,16 @@
...
+    sudo apt-get install          \
+      musl-tools                  \
+      python3-toml
+    cargo install cargo-deb
+    ;;
+  "macos-latest")
+    rustup target add x86_64-apple-darwin
+    pip install toml
+    ;;
+esac
```
tommilligan, line 17: ditto here, have an else clause that errors

tommilligan (changes requested):
Looks great, I checked the GCP bucket and all looks good with the artefacts you uploaded. Couple of nits and then should be good to go

.github/workflows/publish.yml
margold: Yes, there is some bug with using cache on mac. If you think we should look for workarounds until it's fixed in github actions, I can create a new ticket.

tommilligan (approved):
Great, good to go! :macau:

.github/workflows/publish.yml
tommilligan: No, I think it's fine while publishing is infrequent. If this became more a CD thing we could investigate it

PR #56, opened 2021-03-03
api/cli: support for HTTP proxies

mcobzarenco:
Add support for using an HTTP proxy for all requests. The proxy configuration is saved as part of the context.

Additionally, the proxy can be overridden / specified as a one off using a global command line argument `--proxy https://proxy.example` (e.g. similar to `--endpoint` / `--token`).

E.g.
```
re get comments // uses proxy from context
re --proxy http://112.78.165.190:8080 // overwrites the proxy from the context with the given one
````

To add a proxy to a context
```
re config add -n my-context --proxy http://myproxy.example
```

NB. When modifying a context interactively, the user is not asked about changing the proxy because I had to deal with it being an `Option<Option<<>>` and couldn't reuse the stdin input util and didn't bother.

I tried it out with online free HTTP proxies. **Don't use a token / context for a real account if you do this and reset it afterwards anyway** !!

For https://github.com/reinfer/platform/issues/10508

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "reinfer-cli"
-version = "0.5.1"
+version = "0.5.2"
 description = "Command line interface for the reinfer platform."
 homepage = "https://github.com/reinfer/cli"
 readme = "README.md"
@@ -32,8 +32,9 @@ reqwest = { version = "0.11.0", default-features = false }
 serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.55"
 structopt = { version = "0.3.15", default-features = false }
+url = {version= "2.2.1", features = ["serde"] }
 
-reinfer-client = { version = "0.5.1", path = "../api" }
+reinfer-client = { version = "0.5.2", path = "../api" }
 
 [dev-dependencies]
 uuid = { version = "0.8.1", features = ["v4"] }
--- a/cli/src/args.rs
+++ b/cli/src/args.rs
@@ -38,10 +38,15 @@ pub struct Args {
     /// context, if any.
     pub token: Option<String>,
 
+    #[structopt(long = "proxy")]
+    /// URL for an HTTP proxy that will be used for all requests if specified
+    pub proxy: Option<Url>,
+
     #[structopt(subcommand)]
     pub command: Command,
 }
 
+#[allow(clippy::large_enum_variant)]
 #[derive(Debug, StructOpt)]
 pub enum Command {
     #[structopt(name = "completion")]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+# v0.5.2
+
+ - Add support for using an HTTP proxy for all requests. The proxy configuration is saved as part of the context.
+   Additionally, the proxy can be overridden / specified as a one off using a global command line argument
+   `--proxy https://proxy.example` (e.g. similar to `--endpoint`).
+
 # v0.5.1
 
 ## Breaking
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

File: cli/src/args.rs [hunk 1]
```diff
@@ -38,10 +38,15 @@ pub struct Args {
     /// context, if any.
     pub token: Option<String>,
 
+    #[structopt(long = "proxy")]
+    /// URL for an HTTP proxy that will be used for all requests if specified
+    pub proxy: Option<Option<Url>>,
```
mcobzarenco, line 43: a context optionally has a proxy and the argument is itself option, so `Option<Option<Url>>` -- e.g.
```
re get comments // uses proxy from context; proxy == None
re --proxy get comments // overwrites the proxy from the context to use no proxy; proxy == Some(None)
re --proxy http//example.proxy // overwrites the proxy from the context to use http//example.proxy; proxy == Some(Some("http//example.proxy"))
```

cli/Cargo.toml
```diff
@@ -32,6 +32,7 @@ reqwest = { version = "0.11.0", default-features = false }
 serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.55"
 structopt = { version = "0.3.15", default-features = false }
+url = {version= "2.2.1", features = ["serde"] }
 
 reinfer-client = { version = "0.5.1", path = "../api" }
```
margold, line 37: Does the version here need to be incremented also?

margold:
Can't comment on the code, but tested this version locally and it worked as expected (tested without valid API token so it amounted to receiving 401 Unauthorised trying out different subcommands)

mcobzarenco:
Edit: for reasons discussed on [slack](https://reinfer.slack.com/archives/C0APLNE23/p1614792136070600?thread_ts=1614791376.067200&cid=C0APLNE23), it's not possible to remove the proxy as a one-off any more
```
re --proxy get comments // overwrites the proxy from the context to use no proxy
```

mcobzarenco:
I fixed the empty proxy arg I mentioned on Slack as well as the version mismatch Irina pointed out and pushed force

tommilligan (changes requested):
Looks good to me. Reading through your discussion on `--proxy`, I'd say the expected use case of passing `--proxy` to **disable** the proxy is pretty confusing anyway - I think if you wanted a on-off disable, I'd add a `--no-proxy` flag and just validate we don't have both given at once.

As you point out, we no longer support disabling the proxy as a one-off, so we should flatten the `Option<Option<Url>>` to reflect that:

```diff
diff --git a/cli/src/commands/config.rs b/cli/src/commands/config.rs
index 137a420..f5ccb86 100644
--- a/cli/src/commands/config.rs
+++ b/cli/src/commands/config.rs
@@ -35,7 +35,7 @@ pub enum ConfigArgs {

         #[structopt(long = "proxy")]
         /// URL for an HTTP proxy that will be used for all requests if specified
-        proxy: Option<Option<Url>>,
+        proxy: Option<Url>,
     },

     #[structopt(name = "current")]
@@ -174,7 +174,7 @@ fn add_or_edit_context(
     token: &Option<String>,
     endpoint: &Option<Url>,
     accept_invalid_certificates: bool,
-    proxy: &Option<Option<Url>>,
+    proxy: &Option<Url>,
     mut config: ReinferConfig,
     config_path: impl AsRef<Path>,
 ) -> Result<()> {
@@ -242,7 +242,7 @@ fn add_or_edit_context(
         endpoint,
         token,
         accept_invalid_certificates,
-        proxy: proxy.clone().unwrap_or_else(|| {
+        proxy: proxy.clone().or_else(|| {
             existing_context
                 .as_ref()
                 .and_then(|context| context.proxy.clone())
```

mcobzarenco:
@tommilligan Awesome, thanks for the review!

> As you point out, we no longer support disabling the proxy as a one-off, so we should flatten the Option<Option<Url>> to reflect that

Where you point it out it's when editing contexts where it's still possible to remove a proxy by specifying an empty argument.

tommilligan (approved):
:facepalm: in that case, ignore me. Approved!

PR #58, opened 2021-03-05
cli: add bulk delete command

mcobzarenco:
Adds a new CLI command for deleting comments in bulk in a source. 

Example usage to delete all unannotated comments in the (inclusive) range `2015-01-01T00:00:00Z` -> `2017-07-30T00:00:00Z`.
```
re delete bulk --source org/delete-me --from-timestamp 2015-01-01T00:00:00Z --to-timestamp 2017-07-30T00:00:00Z
```
By default, comments that are annotated in any dataset containing the source are NOT deleted. Note that only datasets visible to the user account making the request are considered. To delete annotated comments as well, use the `--include-annotated` flag (has to be `=true`)
```
re delete bulk --source org/delete-me --from-timestamp 2015-01-01T00:00:00Z --to-timestamp 2017-07-30T00:00:00Z --include-annotated=true
```

---

Notes:
  - I called everything "annotations" / "annotated" instead of "reviewed", as that's the new name we're using these days, although "reviewed" is  still used in the CLI when getting comments e.g. `--reviewed-only`
  - We are a bit inconsistent with when a flag can be used by itself without a value, e.g. `--no-progress`  whereas in some places we  need an explicit truth value, e.g. `include-predictions=true`, `--has-sentiment=true`, `--should-translate=false`. Defo prefer the latter, but I added a `--no-progress` flag without a value for consistency, e.g. `re delete bulk --include-annotated=true --no-progress ...`
  
---

Demo:

Peek 2021-03-05 14-44

^ deleting all unreviewed comments followed by deleting all comments ^

For reinfer/platform#9757

### DIFF
--- a/api/src/resources/comment.rs
+++ b/api/src/resources/comment.rs
@@ -120,6 +120,9 @@ pub struct Comment {
     #[serde(skip_serializing_if = "PropertyMap::is_empty", default)]
     pub user_properties: PropertyMap,
     pub created_at: DateTime<Utc>,
+
+    #[serde(default)]
+    pub has_annotations: bool,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
--- a/cli/src/commands/delete.rs
+++ b/cli/src/commands/delete.rs
@@ -1,7 +1,18 @@
 use anyhow::{Context, Result};
-use log::info;
-use reinfer_client::{BucketIdentifier, Client, CommentId, DatasetIdentifier, SourceIdentifier};
+use chrono::{DateTime, Utc};
+use colored::Colorize;
+use std::sync::{
+    atomic::{AtomicUsize, Ordering},
+    Arc,
+};
 use structopt::StructOpt;
+
+use reinfer_client::{
+    BucketIdentifier, Client, CommentId, CommentsIterTimerange, DatasetIdentifier, Source,
+    SourceIdentifier,
+};
+
+use crate::progress::{Options as ProgressOptions, Progress};
 
 #[derive(Debug, StructOpt)]
 pub enum DeleteArgs {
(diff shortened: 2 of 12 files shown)


cli/src/progress.rs
```diff
@@ -14,25 +14,11 @@ pub type ProgressMessage = (u64, String);
 
 pub struct Options {
     pub bytes_units: bool,
-    pub show_bar: bool,
```
mcobzarenco, line 17: were unused

cli/src/progress.rs
```diff
@@ -45,7 +31,7 @@ impl Progress {
     pub fn new<ProgressFnT, StatisticsT>(
         progress_fn: ProgressFnT,
         statistics: &Arc<StatisticsT>,
-        target_value: u64,
```
mcobzarenco, line 48: Passing in `None` means an unbounded progress bar (e.g. spinner) -- we don't have total counts for how many comments are in the range (as it's a per source operation)

tommilligan:
Will leave the code review to Cristi, but I question why we need the `--include-annotated` flag? I would expect `delete bulk` to delete all comments uniformly, and some flag like `--exclude-annotated` to be an exception to this.

From your original description of UBS wanting to delete historical data, they don't need to ignore annotated comments for some reason?

mcobzarenco:
> Will leave the code review to Cristi, but I question why we need the --include-annotated flag? I would expect delete bulk to delete all comments uniformly, and some flag like --exclude-annotated to be an exception to this.

Fwiw, that was UBS' requirement -- e.g. they don't want to ever delete annotated comments. My thinking was that by default the command doesn't delete any data generated by the users.

Cristi brought this up too, he suggested making `--include-annotated` a required argument, that solution seems pretty nice as it's always explicit :+1:

mcobzarenco:
I made `--include-annotated` required and changed the impl to use the new `has_annotations` field returned by the API -- which simplifies the code and means we don't need to get datasets and labellings any more

api/src/resources/comment.rs
```diff
@@ -120,6 +120,9 @@ pub struct Comment {
     #[serde(skip_serializing_if = "PropertyMap::is_empty", default)]
     pub user_properties: PropertyMap,
     pub created_at: DateTime<Utc>,
+
+    #[serde(default = "default_is_false")]
```
cristicbz, line 124: why not `#[serde(default)]`, the `Default` impl is `false` for `bool` so i think this should work

cristicbz (approved):
LGTM!

mcobzarenco:
Awesome -- will have to deploy new API before merging

cli/src/commands/delete.rs
```diff
@@ -48,26 +89,208 @@ pub fn run(delete_args: &DeleteArgs, client: Client) -> Result<()> {
...
+        },
+        include_annotated,
+    );
+
+    let skip_ids = if include_annotated {
+        log::warn!("All matching comments will be deleted, including all annotations.");
+        HashSet::new()
+    } else {
+        let datasets_with_source = client
+            .get_datasets()?
```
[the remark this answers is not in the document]
mcobzarenco: Solved by using a new field on comment objects returned by the API which marks comments with annotations

PR #63, opened 2021-04-19
buckets: make transform tag required

tommilligan:
Relates to https://github.com/reinfer/platform/issues/7012

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,18 @@
 # Unreleased
 
+## Breaking
+
+- The `create bucket` flag `--transform-tag` is now required.
+
 # v0.5.3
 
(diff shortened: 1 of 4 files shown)


AWilcke (changes requested):
Awesome stuff, default transform tags are evil! :imp: 

I see we already have a test that invalid transform tags fail, we should probably have a `test_create_bucket_without_transform_tag_fails` as well or something like that?

tommilligan:
> I see we already have a test that invalid transform tags fail, we should probably have a `test_create_bucket_without_transform_tag_fails` as well or something like that?

We could, but at the moment that's basically just testing `structopt`, so I don't think it's worth it

AWilcke (approved):
Fair enough! :rocket:

PR #68, opened 2021-05-04
release: v0.6.0

margold:


davidhewitt:
:+1: let's get #67 and #46 in just quickly before doing so...

davidhewitt (approved):
I suggest rebasing on `master` to pick up the above PRs, then let's :ship:

PR #69, opened 2021-05-21
Fix inability to remove labellings by posting empty lists

mcobzarenco:
When uploading annotated comments, empty lists of assigned / dismissed labels are now serialized in the request. Previously empty lists were skipped which meant it was not possible to remove labellings (N.B. the API distinguishes between missing field -- labellings are unmodified -- or an empty list -- labellings are removed).


tommilligan (approved):
:receipt:

PR #71, opened 2021-06-16
cli: update dataset/source

tommilligan:
Add cli and api constructs to allow updating sources and datasets.

As well as just increasing coverage of our API, we need this for `gpt-data-generator` internally.

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -28,11 +28,13 @@ use crate::resources::{
     dataset::{
         CreateRequest as CreateDatasetRequest, CreateResponse as CreateDatasetResponse,
         GetAvailableResponse as GetAvailableDatasetsResponse, GetResponse as GetDatasetResponse,
+        UpdateRequest as UpdateDatasetRequest, UpdateResponse as UpdateDatasetResponse,
     },
     email::{PutEmailsRequest, PutEmailsResponse},
     source::{
         CreateRequest as CreateSourceRequest, CreateResponse as CreateSourceResponse,
         GetAvailableResponse as GetAvailableSourcesResponse, GetResponse as GetSourceResponse,
+        UpdateRequest as UpdateSourceRequest, UpdateResponse as UpdateSourceResponse,
     },
     statistics::GetResponse as GetStatisticsResponse,
     trigger::{
@@ -64,12 +66,12 @@ pub use crate::{
         },
         dataset::{
             Dataset, FullName as DatasetFullName, Id as DatasetId, Identifier as DatasetIdentifier,
-            Name as DatasetName, NewDataset,
+            Name as DatasetName, NewDataset, UpdateDataset,
         },
         email::{Id as EmailId, Mailbox, MimeContent, NewEmail},
         source::{
             FullName as SourceFullName, Id as SourceId, Identifier as SourceIdentifier,
-            Name as SourceName, NewSource, Source,
+            Name as SourceName, NewSource, Source, UpdateSource,
         },
         statistics::Statistics,
         trigger::{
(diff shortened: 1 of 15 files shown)


api/src/lib.rs
```diff
@@ -193,6 +195,21 @@ impl Client {
             .source)
     }
 
+    /// Update a new source.
```
AWilcke, line 198: nit: I'm assuming this shouldn't be a `new` source

api/src/resources/dataset.rs
```diff
@@ -146,3 +156,25 @@ pub(crate) struct GetAvailableResponse {
...
+
+#[derive(Debug, Clone, Serialize)]
+pub(crate) struct UpdateRequest<'request> {
+    pub dataset: UpdateDataset<'request>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub(crate) struct UpdateResponse {
+    pub dataset: Dataset,
+}
```
AWilcke, line 180: Are these structs there so that the `serde`-generated `json` matches the `api` schema?

cli/src/commands/create/dataset.rs
```diff
@@ -1,29 +1,15 @@
...
-
-/// Ensures the name provided conforms to the <owner>/<name> structure
-fn validate_dataset_name(s: String) -> Result<(), String> {
-    if !RX_FULL_NAME.is_match(&s) {
-        return Err(s);
-    }
-
-    Ok(())
-}
-
```
AWilcke, line 21: as far as I can tell we're removing this and the only validation left is that it's two parts split by `/`, when using `FullName::from_str`. Do we not want to keep the regex validation?

cli/src/commands/update/dataset.rs
```diff
@@ -0,0 +1,70 @@
...
+                .iter()
+                .map(|source| Ok(client.get_source(source.clone())?.id))
+                .collect()
+        })
+        .transpose()
+        .context("Operation to get sources failed")?;
+
+    let dataset_full_name = match dataset.to_owned() {
+        DatasetIdentifier::FullName(name) => name,
+        dataset @ DatasetIdentifier::Id(_) => client
```
AWilcke, line 47: never seen this use of `@`, here you're binding `dataset` to a `DatasetIdentifer` with id = `dataset.to_owned()` is that right?

because `dataset` can be either a full name (first match case) or an id (second match case)?

cli/src/commands/update/dataset.rs
```diff
@@ -0,0 +1,70 @@
...
+            sources
+                .iter()
+                .map(|source| Ok(client.get_source(source.clone())?.id))
+                .collect()
+        })
+        .transpose()
+        .context("Operation to get sources failed")?;
+
+    let dataset_full_name = match dataset.to_owned() {
+        DatasetIdentifier::FullName(name) => name,
```
AWilcke, line 46: is `from_str` constructor called automatically if the argument is a string?

AWilcke (changes requested):
Mainly rust-questions for my education, but also I don't think you've added any tests for the new commands? Most of the other commands have tests so I think it would be nicer to have some here as well

api/src/resources/dataset.rs
tommilligan: Yes, exactly. We could probably reduce the number of structs we have if we really wanted to (I think `UpdateResponse` and `CreateResponse` are actually the same), but for now I'm happy sticking to the 1:1 mapping until it gets painful for the sake of clarity.

cli/src/commands/create/dataset.rs
tommilligan: No, I don't think we do. We don't validate any other ids with regexes in this way. We leave pretty much all the validation up to the API and report those errors back to the user, which I think is actively good - there's no point reimplementing it all client side.

This change brings it into line with how we talk about source/bucket ids.

cli/src/commands/update/dataset.rs
tommilligan: Yes, pretty much spot on. The `variable @ pattern` syntax allows you to match with a pattern, but not deconstruct with it. This is important in Rust because normally deconstructing takes ownership. Wheras here we want to reuse the data, so the `@` allows us to avoid that.

cli/src/commands/update/dataset.rs
tommilligan: I'm not sure what you mean here? `DatasetIdentifier::FullName(name)` is a pattern that allows us to take the inner value `name` out of the enum variant `DatasetIdentifier::FullName`. There's no construction here.

`from_str` is called automatically by `StructOpt` when parsing the cli arguments into the `args`, at the beginning of the program.

AWilcke (approved):
Awesome stuff :rocket:

PR #73, opened 2021-06-21
api: make resource f64 NotNan, add more Eq/Serialize traits

tommilligan:
Really boring PR that increase utility of the `reinfer_client` create for downstream applications.

- makes all uses of `f64` in our api types `NotNan<f64>`. This allows us to derive a bunch more useful traits, notably `Eq`
  - we already do the same thing in our rust services internally, as we don't allow `NaN`s anywhere in our data model
- derives and (hopefully) standardises a bunch of the traits we derive on our data types. That is:
  - make anything that's id-ish `Hash` such as `Fullname`, `Id` and `Identifier`-type newtypes. This allows downstream use in e.g. `HashMap`
  - make as many of our structs `Debug, Clone, Serialize, Deserialize, PartialEq, Eq` as we feasibly can
    - requests cannot derive `Deserialize` as they are borrowed

I looked to see if there was a better way to standardise multiple derives, but it doesn't seem there is beyond [making a macro](https://stackoverflow.com/questions/56450533/is-there-a-way-to-alias-multiple-derives-as-a-single-one) to do a bunch of derives for us, something like:
```rust
reinfer_id! { pub struct Id(pub String); }
```
Which I don't particularly like and doesn't look very composeable.


davidhewitt (approved):
Changes look fine to me!

> I looked to see if there was a better way to standardise multiple derives, but it doesn't seem there is beyond making a macro to do a bunch of derives for us, something like:

Agreed that a `macro_rules!` macro in this way isn't so nice. It wouldn't be too hard to write a proc macro which could be applied as an attribute (e.g. `#[dataclass_derives]`). I still think it's pretty horrendous.

A slightly nicer way to do it would be to have a macro for testing `assert_standard_derives!(Id)`. It can expand to something like this:

```
fn assert_hash<T: Hash>() {};

macro_rules! assert_standard_derives {
    ($ty:ident) => {
        assert_hash::<$ty>();
    };
}
```

Got to remember to add all the structs to some big test, but at least it can ensure that all the right derives are present. (Types which aren't `Hash` in the above code would fail to compile, for example.)

tommilligan:
That's a nice idea, although we then need a macro to check all the structs are in the test suites as you say :stuck_out_tongue: 

A more robust solution would actually be requiring these types to be `Hash` for some unit test or indeed, method withing the `cli` itself, but then that's a bit of an artificial constructed feature.

I think for now I will leave as is, and we can revisit if this becomes a severe issue in future.

PR #75, opened 2021-07-12
api: internal refactoring of entity/label defs

tommilligan:
Started as a PR to expose the existing `entity_defs` fields on retrieved datasets, ended up being: let's make sure the `Dataset` struct is up to date.

- Moves around a bunch of resources in the `reinfer_api` crate for readability and organisation
- Introduces `LabelDef` and tests creating a dataset with it roundtrips
  - no update support yet (will also need to refactor entity defs for consistency and that's bigger work!)
- Remove `EntityDefs` in favour of a generic struct wrapper which feels really nice, and moves this concern from api -> cli


api/src/resources/entity_def.rs
```diff
@@ -0,0 +1,17 @@
...
+pub struct Name(pub String);
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct EntityDef {
+    pub color: u32,
+    pub id: Id,
+    pub inherits_from: Vec<Id>,
+    pub name: Name,
+    pub title: String,
+    pub trainable: bool,
```
annesylvie, line 16: I see that you added some fields that are returned by the api, but might not always be expected by the api when creating a new `EntityDef` (or updating an existing one).
In particular, when deserialising the json blob passed as argument, some of these fields might be missing, and we don't want to fail on that.

So I would either remove the `id` and `color` fields for now, with a `TODO` to add them back later once we do the nicer refactoring that you have done for `LabelDef`-s, or make these fields optional (and skip serialisation when sending them to the api).

annesylvie (commented):
Looks good, I like the new structure around the `LabelDef`-s!
I have one comment about new fields in `EntityDef` though.

api/src/resources/entity_def.rs
tommilligan: Right, good point. I'll update with an equivalent `NewEntityDef` factoring

annesylvie (approved):
LGTM! Thanks for doing the extra refactoring!

PR #76, opened 2021-07-23
api: add flatten serde field to Label

annesylvie:
This enables us to roundtrip annotation metadata once these are returned by the api.


tommilligan (approved):
Looks good!

PR #82, opened 2021-09-02
Deserialize thread properties when using labellings route

mcobzarenco:
This PR is a hot fix to provide a way to download thread properites. These are now deserialized when dowloading comments for a dataset, `re get comments source -d dataset`. This limitation exists as only the `/labellings` API route returns thread properties.

Note, the thread properties are returned top-level as that's what the API does, e.g. 
```json
{
  "comment": {
    "id": "11047305213",
    "uid": "530f10db824404b5.11047305213",
    "thread_id": "3c3131303437333035323e",
    "timestamp": "2020-01-09T16:34:45Z",
    "messages": [],
    "user_properties": {}
  },
  "thread_properties": {
    "duration": 0,
    "response_time": 0,
    "num_messages": 19,
    "thread_position": 5,
    "first_sender": "cs@farfetch.com"
  }
}
```

### DIFF
--- a/api/src/resources/comment.rs
+++ b/api/src/resources/comment.rs
@@ -339,6 +339,8 @@ pub struct AnnotatedComment {
     pub labelling: Option<Labelling>,
     #[serde(skip_serializing_if = "should_skip_serializing_entities")]
     pub entities: Option<Entities>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub thread_properties: Option<ThreadProperties>,
 }
 
 impl AnnotatedComment {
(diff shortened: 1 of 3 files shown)


api/src/resources/comment.rs
```diff
@@ -114,12 +114,12 @@ pub struct SyncCommentsResponse {
     pub unchanged: usize,
 }
 
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
```
mcobzarenco, line 117: Removing `Eq` because floats in Rust

mcobzarenco: Nevermind this, I noticed we already use `ordered_float` and `NotNan` for this reason.

joe-prosser:
👍

PR #84, opened 2021-09-24
projects: add crud commands

davidhewitt:
This PR adds commands to manage projects via the CLI.

### DIFF
--- a/cli/src/commands/create/project.rs
+++ b/cli/src/commands/create/project.rs
@@ -1,0 +1,47 @@
+use crate::printer::Printer;
+use anyhow::{Context, Result};
+use log::info;
+use reinfer_client::{Client, NewProject, ProjectName, UserId};
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+pub struct CreateProjectArgs {
+    #[structopt(name = "project-name")]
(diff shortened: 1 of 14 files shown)


cli/src/commands/create/project.rs
```diff
@@ -0,0 +1,47 @@
...
+    #[structopt(long = "title")]
+    /// Set the title of the new project
+    title: Option<String>,
+
+    #[structopt(long = "description")]
+    /// Set the description of the new project
+    description: Option<String>,
+
+    #[structopt(long = "user-ids", required = true)]
+    /// Set the description of the new project
```
tommilligan, line 22: nit: bad comment

cli/src/commands/create/project.rs
```diff
@@ -0,0 +1,47 @@
...
+    user_ids: Vec<UserId>,
```
tommilligan, line 23: This could also be usernames I guess in future, but agree user ids are fine for now

tommilligan (approved):
Amazing, thanks for adding this :pray:

mcobzarenco (approved):
Thanks so much for adding these commands, super useful :100:

PR #87, opened 2021-09-24
response: simplify error responses

davidhewitt:
Just a quick cleanup to remove a trait abstraction and use a fixed struct.


tommilligan (approved):
I think there's an argument to be made that the v2 API messages may have a different structure? But we can always implement again if that's the case, agree we don't need multiple error types atm

PR #99, opened 2021-11-03
api: make thread position optional

joe-prosser:
When uploading annotated comments `thread_position` is not guaranteed to be returned by the API. This PR ensures that uploads do not fail when those `thread_position` is not present in the response.

### DIFF
--- a/api/src/resources/comment.rs
+++ b/api/src/resources/comment.rs
@@ -419,7 +419,7 @@ pub struct ThreadProperties {
     duration: Option<NotNan<f64>>,
     response_time: Option<NotNan<f64>>,
     num_messages: u64,
-    thread_position: u64,
+    thread_position: Option<u64>,
     first_sender: Option<String>,
 }
 


api/src/resources/comment.rs
```diff
@@ -418,8 +418,8 @@ impl AnnotatedComment {
 pub struct ThreadProperties {
     duration: Option<NotNan<f64>>,
     response_time: Option<NotNan<f64>>,
-    num_messages: u64,
-    thread_position: u64,
+    num_messages: Option<u64>,
```
dan-ri, line 421:
```suggestion
    num_messages: u64,
```

dan-ri: As discussed, `num_messages` is always present if thread properties are present at all.

dan-ri (changes requested):
Thanks!

dan-ri (approved):

PR #100, opened 2021-11-03
comments: Return id for failed comment.

RobSumner:
Adds a `uid` to the error message when uploading annotations - I found this much more useful when debugging failing annotation uploads.

### DIFF
--- a/cli/src/commands/create/comments.rs
+++ b/cli/src/commands/create/comments.rs
@@ -280,7 +280,9 @@ fn upload_batch(
                     labelling.as_deref(),
                     entities.as_ref(),
                 )
-                .context("Could not update labelling for comment")?;
+                .with_context(|| {
+                    format!("Could not update labelling for comment `{}`", comment_uid.0,)
+                })?;
             statistics.add_annotation();
         }
     }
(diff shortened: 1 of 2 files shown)


cli/src/commands/create/comments.rs
```diff
@@ -280,7 +280,10 @@ fn upload_batch(
                     labelling.as_deref(),
                     entities.as_ref(),
                 )
-                .context("Could not update labelling for comment")?;
+                .context(format!(
+                    "Could not update labelling for comment `{}`",
+                    comment_uid.0,
+                ))?;
```
tommilligan, line 286: You probably want `with_context`, which is lazily evaluated - otherwise you're allocating a new string even if the error case isn't hit.

https://docs.rs/anyhow/1.0.45/anyhow/trait.Context.html#tymethod.with_context

```suggestion
                .with_context(|| {
                    format!("Could not update labelling for comment `{}`", comment_uid.0,)
                })?;
```

tommilligan (approved):
Great idea, just nit on the implementation - don't need to review again.

Also looks like there's an identical case in `create/annotations.rs`, so you may want to add there as well?

RobSumner:
The case in `create/annotations.rs` is slightly different (compiler complains about borrowing `new_comment.comment.id` when it's already partially borrowed) so might need more refactoring. I'll merge this as is and tackle that when I have some free time.

tommilligan:
Yeah no worries. I took a quick peek as I was interested, don't think it was anything particularly fun though so I just adding another tiny PR here: https://github.com/reinfer/cli/pull/101

PR #101, opened 2021-11-08
annotations: return id for failed comment upload

tommilligan:
Follow up from #100 

Also print failed uid on annotation upload, not just comment.


RobSumner (approved):
Much easier than I thought! Thanks for fixing this.

PR #105, opened 2021-11-29
api/cli: add source-type parameter to source create [RE-913]

jcalero:
Add `--source-type` parameter to source creation.

### DIFF
--- a/api/src/resources/source.rs
+++ b/api/src/resources/source.rs
@@ -1,5 +1,6 @@
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
+use serde_with::{DeserializeFromStr, SerializeDisplay};
 use std::{
     fmt::{Display, Formatter, Result as FmtResult},
     str::FromStr,
@@ -22,6 +23,9 @@ pub struct Source {
     pub should_translate: bool,
     pub created_at: DateTime<Utc>,
     pub updated_at: DateTime<Utc>,
+
+    #[serde(rename = "_type")]
+    pub source_type: SourceType,
 }
 
 impl Source {
@@ -104,6 +108,39 @@ impl Display for Identifier {
     }
 }
 
+#[derive(Debug, Clone, SerializeDisplay, DeserializeFromStr, PartialEq, Eq, Hash)]
+pub enum SourceType {
+    Call,
+    Chat,
+    Unknown(Box<str>),
+}
+
+impl FromStr for SourceType {
+    type Err = Error;
+
+    fn from_str(string: &str) -> Result<Self> {
+        Ok(match string {
+            "call" => SourceType::Call,
+            "chat" => SourceType::Chat,
+            value => SourceType::Unknown(value.into()),
+        })
+    }
+}
+
+impl Display for SourceType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                SourceType::Call => "call",
+                SourceType::Chat => "chat",
+                SourceType::Unknown(value) => value.as_ref(),
+            }
+        )
+    }
+}
+
 #[derive(Debug, Clone, Serialize, PartialEq, Eq, Default)]
 pub struct NewSource<'request> {
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -123,6 +160,9 @@ pub struct NewSource<'request> {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub sensitive_properties: Option<Vec<&'request str>>,
+
+    #[serde(skip_serializing_if = "Option::is_none", rename = "_type")]
+    pub source_type: Option<&'request SourceType>,
 }
 
 #[derive(Debug, Clone, Serialize, PartialEq, Eq, Default)]
@@ -171,4 +211,38 @@ pub(crate) struct UpdateRequest<'request> {
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
 pub(crate) struct UpdateResponse {
     pub source: Source,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn source_type_roundtrips() {
+        assert_eq!(SourceType::Call, SourceType::from_str("call").unwrap());
+        assert_eq!(
+            &serde_json::ser::to_string(&SourceType::Call).unwrap(),
+            "\"call\""
+        );
+
+        assert_eq!(SourceType::Chat, SourceType::from_str("chat").unwrap());
+        assert_eq!(
+            &serde_json::ser::to_string(&SourceType::Chat).unwrap(),
+            "\"chat\""
+        );
+    }
+
+    #[test]
+    fn unknown_source_type_roundtrips() {
+        let source_type = SourceType::from_str("unknown").unwrap();
+        match &source_type {
+            SourceType::Unknown(error) => assert_eq!(&**error, "unknown"),
+            _ => panic!("Expected error to be parsed as Unknown(..)"),
+        }
+
+        assert_eq!(
+            &serde_json::ser::to_string(&source_type).unwrap(),
+            "\"unknown\""
+        )
+    }
 }
--- a/cli/src/commands/create/source.rs
+++ b/cli/src/commands/create/source.rs
@@ -1,7 +1,7 @@
 use crate::printer::Printer;
 use anyhow::{Context, Result};
 use log::info;
-use reinfer_client::{BucketIdentifier, Client, NewSource, SourceFullName};
+use reinfer_client::{BucketIdentifier, Client, NewSource, SourceFullName, SourceType};
 use structopt::StructOpt;
 
 #[derive(Debug, StructOpt)]
@@ -29,6 +29,10 @@ pub struct CreateSourceArgs {
     #[structopt(long = "bucket")]
     /// Bucket to pull emails from.
     bucket: Option<BucketIdentifier>,
+
+    #[structopt(long = "source-type")]
+    /// Set the source type of the new source
+    source_type: Option<SourceType>,
 }
 
 pub fn create(client: &Client, args: &CreateSourceArgs, printer: &Printer) -> Result<()> {
@@ -39,6 +43,7 @@ pub fn create(client: &Client, args: &CreateSourceArgs, printer: &Printer) -> Re
         language,
         should_translate,
         bucket,
+        source_type,
     } = args;
 
     let bucket_id = match bucket.to_owned() {
@@ -62,6 +67,7 @@ pub fn create(client: &Client, args: &CreateSourceArgs, printer: &Printer) -> Re
                 should_translate: *should_translate,
                 bucket_id,
                 sensitive_properties: None,
+                source_type: source_type.as_ref(),
             },
         )
         .context("Operation to create a source has failed")?;
--- a/cli/tests/test_sources.rs
+++ b/cli/tests/test_sources.rs
@@ -119,6 +119,7 @@ fn test_create_update_source_custom() {
         description: String,
         language: String,
         should_translate: bool,
+        source_type: String,
     }
 
     impl From<Source> for SourceInfo {
@@ -130,6 +131,7 @@ fn test_create_update_source_custom() {
                 description: source.description,
                 language: source.language,
                 should_translate: source.should_translate,
+                source_type: source.source_type.to_string(),
             }
         }
     }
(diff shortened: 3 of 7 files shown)


api/src/resources/source.rs
```diff
@@ -123,6 +124,9 @@ pub struct NewSource<'request> {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub sensitive_properties: Option<Vec<&'request str>>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub _type: Option<&'request str>,
```
mcobzarenco, line 129: nit. While, I agree with this being called `_type` in the API, I am not so sure whether this should extend to client / cli as 1. this has no notion of v1. vs v2., the source here is just an abstract resource and more pragmatically, `_field` in Rust by convention is an unused field (i.e. it would not warn if you don't use it, e.g. as a tag or for FFI). `type` is a keyword though, my nit suggestion would be `type_`

api/src/resources/source.rs
```diff
@@ -22,6 +22,7 @@ pub struct Source {
     pub should_translate: bool,
     pub created_at: DateTime<Utc>,
     pub updated_at: DateTime<Utc>,
+    pub _type: String,
```
mcobzarenco, line 25: 1. The source type is one of a few cases, so could really be an `enum`. One note of caution -- in the past enums here were a maintenance burder in some places where the list would change often (as such entities or permissions). `String` is a way to make it backwards compatible -- we can add new source types and the cli won't break if it doesn't know about that enum variant. At the same time, I don't expect the list to change often, so if it was me, I'd go for the enum -- also a more challenging change :wink:. 
If not enum, at a minimum, I think we need a new type here similar to id / owner -- can check how those are implemented, you'd have to `impl FromStr for SourceType` so you can take it directly as a cmd line arg.
1. nit. While, I agree with this being called _type in the API, I am not so sure whether this should extend to client / cli as 1. this has no notion of v1. vs v2., the source here is just an abstract resource and more pragmatically, _field in Rust by convention is an unused field (i.e. it would not warn if you don't use it, e.g. as a tag or for FFI). type is a keyword though, my nit suggestion would be type_

mcobzarenco (changes requested):
Looks good -- some small style comments, esp. as for rust practice.

I also think you should add a test that creates a source with a non unknown source type.

api/src/resources/source.rs
jcalero: 1. Sure, I'll give it a go.
2. I completely agree, that's why I have in the backend, I only wanted the think layer of the api serialization to read it as `_type`, wasn't sure it was possible here, but I'll give it a go!

jcalero: Actually, I think I will need help on 2.. I don't get how to do it. How do I map the json key `_type` to the `Source` object, there's no obvious construction of a `Source` object anywhere I can see.

jcalero: As far as I can tell, the cli will call the api and map the api response to a `ResponseT`, which in the case of creating a source is `CreateResponse`, which is just `source: Source`. So the mapping is directly from the api to the `Source` object, hence I don't see how I can name the type field `Source.source_type` rather than `Source._type`. Sounds to me `Source` is just a rust representation of the api json response.

jcalero: Ok, so this worked, is it standard practice?
```
    #[serde(rename = "_type")]
    pub source_type: String,
```

jcalero: On the topic of 1., leaving aside the fun challenge of learning how to do is as an `Enum`, I'm thinking it might not be the smart choice at the moment given we're just introducing this source type and it might change a few times before it settles into a stable type. Having the `create source` cli action break each time we change the api might be something we'll regret!

I'm also not convinced if philosophically the cli should be strict on types when the api is validating. It kinda isn't the cli's responsiblity. It doesn't hurt to help the user, but I think the validation on the cli should always be more relaxed than the validation at the api. It would be pretty obvious for the user when the request is made what they've done wrong anyway.

jcalero:
Updated with what I _think_ resolves your comments, let me know how off the mark I am :sweat_smile:

jcalero:
Oops, actually, forgot to update the tests. Brb!

jcalero:
Ok, updated again. I changed the existing test, since [this PR](https://github.com/reinfer/platform/pull/13044) made it impossible to create a source with `unknown` type anyway, so the only options to test are:
* Setting the type to `call` (or `chat`)
* Omitting the type parameter completely.

api/src/resources/source.rs
```diff
@@ -105,6 +107,17 @@ impl Display for Identifier {
     }
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
+pub struct SourceType(pub String);
+
+impl FromStr for SourceType {
+    type Err = Error;
```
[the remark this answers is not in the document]
mcobzarenco: This implements the trait [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) for you new type `SourceType`. A trait is like an interface -- in this case `FromStr` is a standard library trait used to specify how to get a value of your type from a `String`. 

This is used by the command line library we use (`structopt`) to parse a cmd line argument directly to a `SourceType`.

cli/src/commands/create/source.rs
```diff
@@ -29,6 +29,10 @@ pub struct CreateSourceArgs {
     #[structopt(long = "bucket")]
     /// Bucket to pull emails from.
     bucket: Option<BucketIdentifier>,
+
+    #[structopt(long = "source-type")]
+    /// Set the source type of the new source
+    source_type: Option<String>,
```
mcobzarenco, line 35: This should now be
```suggestion
    source_type: Option<SourceType>,
```
As mentioned above, `structopt` would call `from_str` on `SourceType` to parse a value from the string you provide as an argument.

Currentlym I don't think you're currently using `SourceType`.

cli/src/commands/create/source.rs
```diff
@@ -62,6 +67,7 @@ pub fn create(client: &Client, args: &CreateSourceArgs, printer: &Printer) -> Re
                 should_translate: *should_translate,
                 bucket_id,
                 sensitive_properties: None,
+                source_type: source_type.as_deref(),
```
mcobzarenco, line 70: With the changes above, you'd need to call `.as_ref()` on `source_type` here to go from 
`Option<SourceType>` to `Option<& SourceType>`
```suggestion
                source_type: source_type.as_ref(),
```

cli/tests/test_sources.rs
```diff
@@ -108,6 +108,7 @@ fn test_create_update_source_custom() {
         "--description=some description",
         "--language=de",
         "--should-translate=true",
+        "--source-type=call",
```
mcobzarenco, line 111: If we just modify this test, we don't test creating a source without specifying the type, would it be possible to have a new test that just checks creating a source with `--source-type call` and then source works

api/src/resources/source.rs
```diff
@@ -123,6 +137,9 @@ pub struct NewSource<'request> {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub sensitive_properties: Option<Vec<&'request str>>,
+
+    #[serde(skip_serializing_if = "Option::is_none", rename = "_type")]
+    pub source_type: Option<&'request str>,
```
mcobzarenco, line 142: :+1: on `rename`

I think you should use the newtype here though, e.g. like `BucketId` -- although, you'd want to borrow it rather than copy it 
```suggestion
    pub source_type: Option<&'request SourceType>,
```

api/src/resources/source.rs
mcobzarenco: > On the topic of 1., leaving aside the fun challenge of learning how to do is as an Enum, I'm thinking it might not be the smart choice at the moment given we're just introducing this source type and it might change a few times before it settles into a stable type. Having the create source cli action break each time we change the api might be something we'll regret!

Newtype for string as you've done is not too bad -- but will just mention the solution for permissions was an enum which is either a known variant or an unknown one explicitly. Something like

```rust
#[serde(untagged)]
enum SourceType {
    #[serde(rename = "call")]
    Call,

    #[serde(rename = "chat")]
    Chat,

    Unknown(Box<str>),
}
```

I.e. it's either a known variant or `Unknown(...)` with a string, (`Box<str>` is just a way of saying the string is immutable). You can rely on `serde`'s json parsing for parsing it, e.g. something like (that's what we seem to do for permissions at least)

```rust
impl FromStr for SourceTyoe {
    type Err = Error;

        serde_json::de::from_str(&format!("\"{}\"", string)).map_err(|_| {
            Error::BadSourceType {
                indentifier: string.into(),
            }
        })
    }
}
```

This introduces a new error type `Error::BadSourceType` (see `api/src/error.rs`)

cli/src/commands/create/source.rs
jcalero: I am using `SourceType` I promise! I even had a test failing cause I had implemented it wrong at first. But yeah, I forgot here, oops! Will fix.

cli/tests/test_sources.rs
jcalero: Yeah, that's fair, but we only have one test with all these conditions. We don't have a test without specifying `language`, `description`, `should-translate` or any of the parameters either.

Not saying that's a good thing, so happy to add one, I'm just also aware that it would be a copy and paste of this test pretty much verbatim. If there is a better way I don't think I'm capable of writing it in reasonable time and would need help.

api/src/resources/source.rs
jcalero: Ok!

cli/src/commands/create/source.rs
jcalero: Right :eyes:, I don't really get what's going on here :see_no_evil:

api/src/resources/source.rs
jcalero: That's fair @mcobzarenco , I guess my worry is more about the fact that we ship the CLI, and we'll probably change the api of this type a few times fairly soon. I don't want our clients to have to constantly update the CLI with new updates, or worse yet, have to wait for a fix, whenever we change the `_type` field in the api, especially since it's a private, and currently unstable field.

I'm more than happy to make it a robust feature of the CLI once it's stable, but I don't really think it is at the moment.

api/src/resources/source.rs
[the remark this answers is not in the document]
jcalero: Thanks! But my question was specifically about the line `type Err = Error;`, looks a bit hacky to my eyes, but maybe it's standard practice?

jcalero:
Update once more @mcobzarenco, added a test and updated type references that were missed.

api/src/resources/source.rs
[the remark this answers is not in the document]
mcobzarenco: @jcalero Oh, I see --- that is an [associated type](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html). A trait specifies a bunch of functions you need to implment as well as potentially some types you need to provide. `FromStr` is defined as 

```rust
pub trait FromStr {
    type Err;
    fn from_str(s: &str) -> Result<Self, Self::Err>;
}
```

I.e. if you implement it, you need to specify an error type `Err` and a function `from_str` that returns either
 - a value of type `Self` if the parsing succeeded
 - an error of type `Err` if the parsing failed

That is expressed as returning  `Result<Self, Self::Err>`; btw, `Result<T, E>` is an enum which is either `T` or `E`), e.g.

```
enum Result<T, E> {
  Ok(T),
  Err(E),
}
```

api/src/resources/source.rs
mcobzarenco: > I don't want our users to have to constantly update the CLI with new updates, or worse yet, have to wait for a fix, whenever we change the _type field in the api, especially since it's a private, and currently unstable field.

Just for clarity, the solution above would make it forwards compatible as new `source_type`s introduced by the API would be parsed in the `Unknown` variant of the enum

mcobzarenco (approved):
LGTM -- thanks for all the changes!

api/src/resources/source.rs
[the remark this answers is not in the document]
jcalero: Thanks, I think I get it a bit better now!

jcalero:
@cristicbz is making a case that this isn't going to work either, I'm going to have to write a failing test for this and try again, as my tests pass but I think Cristi is right. So don't bother reviewing yet @mcobzarenco , I'll let you know.

jcalero:
@mcobzarenco , ready for review (again!) :sweat_smile:

mcobzarenco:
Looks great! And really nice catching the bug with permissions 😅 Thank you for documenting it 

One possible supernit just for the sake of learning is that in the `Display` impl you could avoid creating a new `String` (and a memory allocation) as `write!` just need a `&str`, so no need to call `into()`

mcobzarenco (approved):

PR #107, opened 2021-12-03
api/cli: default `has-sentiment` to false

margold:
Don't know if this is controversial, but I think if we make `--has-sentiment` optional then it should default to `false`, since that's what we want most of the time. (I think that it defaults to `true` in api v1 for purely historical reasons). Motivated by accidentally creating sentiment-enabled datasets at UBS multiple times, which was an ordeal to change.

I have no idea what I'm doing here but it seems to work!

### DIFF
--- a/cli/src/commands/create/dataset.rs
+++ b/cli/src/commands/create/dataset.rs
@@ -22,7 +22,10 @@ pub struct CreateDatasetArgs {
     /// Set the description of the new dataset
     description: Option<String>,
 
-    #[structopt(long = "has-sentiment")]
+    #[structopt(
+        long = "has-sentiment",
+        help = "Enable sentiment prediction for the dataset [default: false]"
+    )]
     /// Enable sentiment prediction for the dataset
     has_sentiment: Option<bool>,
 
@@ -95,7 +98,7 @@ pub fn create(client: &Client, args: &CreateDatasetArgs, printer: &Printer) -> R
                 source_ids: &source_ids,
                 title: title.as_deref(),
                 description: description.as_deref(),
-                has_sentiment: *has_sentiment,
+                has_sentiment: Some(has_sentiment.unwrap_or(false)),
                 entity_defs: if entity_defs.is_empty() {
                     None
                 } else {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Breaking
+
+- `re create dataset` will default to sentiment disabled if `--has-sentiment` is not provided.
+
 # v0.10.2
 
 ## Bug Fixes


cli/src/commands/create/dataset.rs
```diff
@@ -22,9 +22,9 @@ pub struct CreateDatasetArgs {
     /// Set the description of the new dataset
     description: Option<String>,
 
-    #[structopt(long = "has-sentiment")]
+    #[structopt(long = "has-sentiment", parse(try_from_str), default_value = "false")]
     /// Enable sentiment prediction for the dataset
-    has_sentiment: Option<bool>,
+    has_sentiment: bool,
```
tommilligan, line 27: Likewise, these types should accurately reflect the user's input. I think we can leave the type how it is, to capture whether the user specified the flag, or left it as the default value.

api/src/resources/dataset.rs
```diff
@@ -122,8 +122,7 @@ pub struct NewDataset<'request> {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub description: Option<&'request str>,
 
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub has_sentiment: Option<bool>,
+    pub has_sentiment: bool,
```
tommilligan, line 125: So the reason this is `Optional` is that these structs (in `api`) should accurately reflect the type of valid requests we can send to our API. In other words, maybe one day we could autogenerate these from openapi :wink: 

This shouldn't change, as our api defs aren't changing

tommilligan (changes requested):
I would implement this with the following change - keeping the input/output types the same, and just changing the logic in the middle:

```diff
diff --git a/cli/src/commands/create/dataset.rs b/cli/src/commands/create/dataset.rs
index 9f35781..6a171ef 100644
--- a/cli/src/commands/create/dataset.rs
+++ b/cli/src/commands/create/dataset.rs
@@ -95,7 +95,7 @@ pub fn create(client: &Client, args: &CreateDatasetArgs, printer: &Printer) -> R
                 source_ids: &source_ids,
                 title: title.as_deref(),
                 description: description.as_deref(),
-                has_sentiment: *has_sentiment,
+                has_sentiment: Some(has_sentiment.unwrap_or(false)),
                 entity_defs: if entity_defs.is_empty() {
                     None
                 } else {
```

You should also update the changelog to note this as a breaking change in behaviour for the cli.

margold:
Thanks, makes sense! I had to manually specify the "help" string since it's not aware of the default value - not sure if there's a better way of doing it.

tommilligan (approved):
Thanks! Yeah I think manual help string is fine. Although maybe that is an argument for using the default value annotation like you were planning?

joe-prosser (approved):

PR #108, opened 2021-12-06
ci: only upload binary when it does not already exist

joe-prosser:
Only upload a binary if it does not already exist - previously this would cause failures when retrying a previous attempt that had failed to publish all required binaries


tommilligan (approved):
LGTM and cool to learn about the `stat` command - thanks for fixing this!

The `latest` files will never be updated with this logic, but it's better than a crash

PR #115, opened 2022-01-18
api/cli: rename source type to kind

jcalero:
This has changed in the api so changing it here for consistency and to support the source response correctly.


joe-prosser (approved):
LGTM ✅  
Thanks!

PR #117, opened 2022-01-19
github: fix publish workflow to use projects env var

jcalero:
Someone forgot to change the publish workflow when he changed the variable name :speak_no_evil:


joe-prosser (approved):
😂

PR #121, opened 2022-01-20
create: trigger-exception

tommilligan:
I needed to test trigger exceptions for my filtering refactoring, so thought I'd add to the CLI rather than working out the curl by hand.

Implemented using the :sparkles: amazing :sparkles: API docs here, as I've never touched this code before: https://developers.reinfer.io/api/reference/triggers#tag-an-exception

```log
$ re create trigger-exception --trigger org2/ee-twitter-nps/hello-integration --type "Sad Violin" --uid "64904092030935f4.2211b04b1b6391172965b57e37d5a9cb"
W TLS certificate verification is disabled. Do NOT use this over an insecure network.
I New trigger exception created successfully
```

### DIFF
--- a/api/src/resources/comment.rs
+++ b/api/src/resources/comment.rs
@@ -36,6 +36,14 @@ impl FromStr for Id {
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
 pub struct Uid(pub String);
 
+impl FromStr for Uid {
+    type Err = Error;
+
+    fn from_str(string: &str) -> Result<Self> {
+        Ok(Self(string.to_owned()))
+    }
+}
+
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
 pub struct ThreadId(pub String);
 
--- a/cli/src/commands/create/trigger_exception.rs
+++ b/cli/src/commands/create/trigger_exception.rs
@@ -1,0 +1,46 @@
+use crate::printer::Printer;
+use anyhow::{Context, Result};
+use log::info;
+use reinfer_client::{
+    Client, CommentUid, TriggerException, TriggerExceptionMetadata, TriggerFullName,
+};
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+pub struct CreateTriggerExceptionArgs {
+    #[structopt(short = "t", long = "trigger")]
+    /// The dataset name or id
+    trigger: TriggerFullName,
+
+    #[structopt(long = "type")]
+    /// The type of exception. Please choose a short, easy-to-understand string such as "No Prediction".
+    r#type: String,
+
+    #[structopt(long = "uid")]
+    /// The uid of the comment that should be tagged as an exception.
+    uid: CommentUid,
+}
+
+pub fn create(
+    client: &Client,
+    args: &CreateTriggerExceptionArgs,
+    _printer: &Printer,
+) -> Result<()> {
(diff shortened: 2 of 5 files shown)


api/src/resources/comment.rs
```diff
@@ -36,6 +36,14 @@ impl FromStr for Id {
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
 pub struct Uid(pub String);
 
+impl FromStr for Uid {
+    type Err = Error;
```
scallaway, line 40: Why do you need to declare this type here? I'm just struggling to see where this is used

cli/src/commands/create/trigger_exception.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[derive(Debug, StructOpt)]
+pub struct CreateTriggerExceptionArgs {
+    #[structopt(short = "t", long = "trigger")]
+    /// The dataset name or id
+    trigger: TriggerFullName,
+
+    #[structopt(long = "type")]
+    /// The type of exception. Please choose a short, easy-to-understand string such as "No Prediction".
+    r#type: String,
```
scallaway, line 17: What's the reason for using a raw string here and not something like `&str`? I'm guessing it's because the values that we're passing in aren't going to be around long enough, so we have to take ownership?

scallaway (approved):
I think I just about follow what's going on in here, but I'm not very familiar with the CLI codebase. I do have a couple of questions just for my understanding inline but happy to approve this as-is.

api/src/resources/comment.rs
tommilligan: This `type` is part of the `FromStr` trait that we're implementing. It is required that we specify what the error type is that this trait returns, which in this case is our global error type, `Error` (the return type `Result<Self>` is an alias for `std::result::Result<Self, Error>`).

More generally, generic traits can have associated types that are named, rather than positional. So in this case, the trait definition is:

```rust
pub trait FromStr {
    type Err;  // <- says the trait implemention must specify this type
    fn from_str(s: &str) -> Result<Self, Self::Err>;
}
```

which you could have defined using positional rather than named associated types, something like:

```rust
pub trait FromStr<Err> {
    fn from_str(s: &str) -> Result<Self, Err>;
}
```

https://doc.rust-lang.org/std/str/trait.FromStr.html#associated-types

cli/src/commands/create/trigger_exception.rs
tommilligan: There's a technical answer about having to add lifetimes to any borrowed values in structs, which structopt doesn't support AFAIK, but the question really boils down to - who owns the data? When I input a `--type "Sad Violin"` flag, who owns `"Sad Violin"`?

You can see the argument parsing call here, right at the start of the program:
https://github.com/reinfer/cli/blob/e75c12914a05d0f371c9207434211e23ecc04e62/cli/src/main.rs#L148

The output is the variable `args`, and **there are no other variables defined before it**. So it would be hard to reference data that's owned by another variable! We could be referencing data in the binary, like a `static`, but that's not the case. We could be referencing data backed by an external resource, but again that's not the case. When we parse the CLI args, we consume them and spit out `args`. So, it **must own all the data**.

That's why we have to use an owned value, like `String`. (I'm not sure what you mean by 'raw string'? The `r#...` syntax is a [raw identifier](https://doc.rust-lang.org/reference/identifiers.html#raw-identifiers) that's used when the identifier would otherwise be a reserved Rust keyword).

PR #124, opened 2022-02-11
cli: show bucket name on get source

margold:
The idea is to show a bucket name if the source is connected to a bucket. In cases where the user can see the source but not the bucket (either bc they don't have permissions to see the bucket, or bc the bucket doesn't exist - example `org1/riley-testuser-mariusreinfer-onmicrosoft-com` on Staging) the bucket ID is shown instead.

### DIFF
--- a/cli/src/commands/get.rs
+++ b/cli/src/commands/get.rs
@@ -1,5 +1,5 @@
 use crate::{
-    printer::{print_resources_as_json, Printer},
+    printer::{print_resources_as_json, EnrichedSource, Printer},
     progress::{Options as ProgressOptions, Progress},
 };
 use anyhow::{anyhow, Context, Result};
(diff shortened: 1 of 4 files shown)


cli/src/commands/get.rs
```diff
@@ -176,7 +176,16 @@ pub fn run(get_args: &GetArgs, client: Client, printer: &Printer) -> Result<()>
                 });
                 sources
             };
-            printer.print_resources(&sources)?;
+            let mut enriched_sources: Vec<EnrichedSource> = Vec::with_capacity(sources.len());
+            for source in sources.into_iter() {
+                let bucket = match source.bucket_id.as_ref() {
+                    // User may not have permission to fetch bucket, or bucket may not exist
+                    Some(bucket_id) => client.get_bucket(bucket_id.clone()).ok(),
```
dan-ri, line 183: Here we're making a separate API call for every source, which is not ideal. Instead we can use `client.get_buckets()` to produce a `HashMap<BucketId, Bucket>` and use that to associate buckets with sources.

cli/src/commands/get.rs
```diff
@@ -176,7 +176,16 @@ pub fn run(get_args: &GetArgs, client: Client, printer: &Printer) -> Result<()>
...
+                    None => None,
+                };
+                enriched_sources.push(EnrichedSource { source, bucket });
```
dan-ri, line 186: As a rule, constructing collections via mutation is an anti-pattern in Rust, and rarely necessary. The appearance of a `let mut` should always arouse suspicion. This can be rewritten without mutation as, say:
```rust
let enriched_sources: Vec<EnrichedSource> = sources
    .into_iter()
    .map(|source| {
        let bucket = source
            .bucket_id
            .as_ref()
            .and_then(|id| client.get_bucket(id.clone()).ok());
        EnrichedSource { source, bucket }
    })
    .collect();
```

dan-ri (changes requested):
Nice job getting this working! The main change we should consider is fetching buckets in a single call instead of a call per source.

PR #125, opened 2022-02-24
user: support get/delete single user

cristicbz:


api/src/resources/user.rs
```diff
@@ -62,6 +62,18 @@ pub enum Identifier {
...
+    type Err = Error;
+
+    fn from_str(string: &str) -> Result<Self> {
+        if string.chars().all(|c| c.is_digit(16)) {
+            Ok(Identifier::Id(Id(string.into())))
+        } else {
+            Username::from_str(string).map(Identifier::Username)
+        }
+    }
+}
```
mcobzarenco, line 75: This is bug prone and surprising imo, e.g. `ada` is a user id; for other resources, e.g. like sources we can figure out if it's an id or full name based on whether there's a slash in the name. 

Can we use a different convention like `id:ada` vs `ada` or something else that uniquely identifies if something is a username or id?

mcobzarenco (changes requested):

api/src/resources/user.rs
cristicbz: You're right. Usernames didn't really work properly anyway, since there is no `get-by-name` route, so I removed username support entirely, can only do id-s now.

api/src/lib.rs
```diff
@@ -232,6 +247,15 @@ impl Client {
         self.delete::<_>(self.endpoints.source_by_id(&source_id)?)
     }
 
+    /// Delete a source.
+    pub fn delete_user(&self, user: impl Into<UserIdentifier>) -> Result<()> {
+        let user_id = match user.into() {
+            UserIdentifier::Id(user_id) => user_id,
+            user @ UserIdentifier::Username(_) => self.get_user(user)?.id,
+        };
+        self.delete::<_>(self.endpoints.user_by_id(&user_id)?)
```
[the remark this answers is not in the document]
cristicbz: There are a lot of unnecessary turbofishes in this file actually. I removed all of them in a separate commite.

mcobzarenco (approved):
LGTM

PR #128, opened 2022-03-25
add project id

tommilligan:
Future support for showing project ids in `re get projects`.

This is not blocked on the corresponding api PR - it will work with current platform, and future versions where id is returned.

### DIFF
--- a/api/src/resources/project.rs
+++ b/api/src/resources/project.rs
@@ -27,8 +27,17 @@ impl FromStr for ProjectName {
     }
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
+pub struct Id(pub String);
+
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
 pub struct Project {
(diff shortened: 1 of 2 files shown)


margold (approved):
Thanks!

api/src/resources/project.rs
```diff
@@ -27,8 +27,17 @@ impl FromStr for ProjectName {
...
+pub struct Id(pub String);
+
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
 pub struct Project {
+    // CODE HEALTH: At time of addition, deployed versions of reinfer may not
+    // return project ids.
+    //
+    // This can be removed once reinfer is deployed everywhere.
+    #[serde(default)]
+    pub id: Option<Id>,
```
mcobzarenco, line 40: Do you know if we can remove the option now?

tommilligan: I doubt it - I don't think we've deployed at UBS in the last 11 days.

mcobzarenco: Ok, cool, I'll merge this if you have no objections

PR #129, opened 2022-04-08
cli: refactor get commands into a module

mcobzarenco:
This PR refactors all the get commands into a module with one file per resource, similar to the `create` commands.


tommilligan (approved):
LGTM, definitely nicer to split smaller

PR #134, opened 2022-06-09
source: support setting transform tag on creation (RE-1978)

margold:
[RE-1978](https://reinfer.atlassian.net/browse/RE-1978)

tested locally after building `api` from https://github.com/reinfer/platform/pull/14729

### DIFF
--- a/api/src/resources/source.rs
+++ b/api/src/resources/source.rs
@@ -9,6 +9,7 @@ use std::{
 use crate::{
     error::{Error, Result},
     resources::bucket::Id as BucketId,
+    resources::bucket::TransformTag,
     resources::user::Username,
 };
 
@@ -26,6 +27,8 @@ pub struct Source {
 
     #[serde(rename = "_kind")]
     pub kind: SourceKind,
+    #[serde(default, rename = "email_transform_tag")]
+    pub transform_tag: Option<TransformTag>,
 }
 
 impl Source {
@@ -163,6 +166,12 @@ pub struct NewSource<'request> {
 
     #[serde(skip_serializing_if = "Option::is_none", rename = "_kind")]
     pub kind: Option<&'request SourceKind>,
+
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        rename = "email_transform_tag"
+    )]
+    pub transform_tag: Option<&'request TransformTag>,
 }
 
 #[derive(Debug, Clone, Serialize, PartialEq, Eq, Default)]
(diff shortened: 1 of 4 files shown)


api/src/resources/source.rs
```diff
@@ -26,6 +27,8 @@ pub struct Source {
 
     #[serde(rename = "_kind")]
     pub kind: SourceKind,
+    #[serde(default, rename = "email_transform_tag")]
+    pub transform_tag: Option<TransformTag>,
```
dan-ri, line 31: Maybe just rename this field to `email_transform_tag` on these objects, since that's what you called it on sources anyway.

cli/tests/test_sources.rs
```diff
@@ -220,6 +220,69 @@ fn test_create_source_with_kind() {
...
+
+    impl From<Source> for SourceInfo {
+        fn from(source: Source) -> SourceInfo {
+            SourceInfo {
+                owner: source.owner.0,
+                name: source.name.0,
+                transform_tag: match source.transform_tag {
+                    Some(transform_tag) => Some(transform_tag.0),
+                    None => None,
+                },
```
dan-ri, line 244:
```suggestion
                transform_tag: source.transform_tag.map(|tag| tag.0),
```

cli/src/printer.rs
```diff
@@ -115,7 +115,7 @@ impl DisplayTable for Project {
 
 impl DisplayTable for Source {
     fn to_table_headers() -> Row {
-        row![bFg => "Name", "ID", "Updated (UTC)", "Title"]
+        row![bFg => "Name", "ID", "Updated (UTC)", "Title", "Bucket"]
```
dan-ri, line 118: Shouldn't this be `"Transform Tag"`? Maybe a bucket field is needed too? May also be best to leave `Title` at the end since it tends to be long.

cli/src/printer.rs
```diff
@@ -124,7 +124,11 @@ impl DisplayTable for Source {
             full_name,
             self.id.0,
             self.updated_at.format("%Y-%m-%d %H:%M:%S"),
-            self.title
+            self.title,
+            match &self.transform_tag {
+                Some(transform_tag) => transform_tag.0.as_str().into(),
+                None => "missing".dimmed(),
+            }
```
dan-ri, line 131:
```suggestion
            self.transform_tag.as_ref().map(|tag| tag.0.as_str().into())
                .unwrap_or_else(|| "missing".dimmed())
```

dan-ri (approved):

cli/src/printer.rs
margold: Yep you're right

PR #138, opened 2022-07-06
github: update config file for publishing action

annesylvie:
By the look of it, `google-github-actions/setup-gcloud` have changed `master` to `main`, updating the config to fix this.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,7 +36,7 @@ jobs:
         run: ./scripts/publish-install
         env:
           BUILD_PLATFORM: ${{ matrix.os }}
-      - uses: google-github-actions/setup-gcloud@master
+      - uses: google-github-actions/setup-gcloud@main
         with:
           service_account_key: ${{ secrets.GCP_SA_KEY }}
       - name: Publish Binaries


tommilligan:
Late to the party but - do we need to use `main`? The reason this broke is that we aren't pinned to a version number, e.g. `v0.6.0` which would remain static over time.

annesylvie:
I don't really have an opinion on this, to be honest I didn't think of pinning a specific version to solve the problem and instead just updated the branch name. If you think pinning to a specific version is nice, I'm happy to do that. Otherwise I think until we have an actual issue with the `google-github-actions/setup-gcloud` repo, I'd rather keep it as it is.

tommilligan:
> I don't really have an opinion on this, to be honest I didn't think of pinning a specific version to solve the problem and instead just updated the branch name. If you think pinning to a specific version is nice, I'm happy to do that. Otherwise I think until we have an actual issue with the `google-github-actions/setup-gcloud` repo, I'd rather keep it as it is.

Happy as github actions isn't on our critical path - for our actual CI, we favour pinning at all times to keep spurious breakage low (we have enough dependencies that otherwise _something_ would be broken every night)

PR #140, opened 2022-07-13
Make clippy happy

margold:
Clippy is failing on master, I think due to a rust update? Attempting to fix.

### DIFF
--- a/cli/src/progress.rs
+++ b/cli/src/progress.rs
@@ -1,4 +1,5 @@
 use indicatif::{ProgressBar, ProgressStyle};
+use std::fmt::Write;
 use std::{
     sync::{
         atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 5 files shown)


cli/src/progress.rs
```diff
@@ -80,14 +81,18 @@ where
 {
     use std::ops::Deref;
     let mut template_str = String::new();
-    template_str.push_str(&format!("{} ", LOG_PREFIX_INFO.deref()));
```
margold, line 83: this was the specific lint: https://rust-lang.github.io/rust-clippy/master/index.html#format_push_string
not sure if this is the best way to fix it

dan-ri: This lint is problematic because of the way the proposed fix introduces fallibility (see [discussion and restriction](https://github.com/rust-lang/rust-clippy/issues/9077)). Still, this change should be fine here.

PR #151, opened 2022-11-04
api/cli: upgrade dependencies and rustc to 1.65.0

mcobzarenco:
Does two things
 - Upgrade all dependencies by running `cargo upgrade` and `cargo update` -- except `indicatif` which was kept back as it needed more changes
 - Fixes clippy lints from rustc 1.65.0


tommilligan (approved):
LGTM, thanks for bringing this up to date!

PR #152, opened 2022-11-10
comments: enable get predictions by timerange

joe-prosser:
Attempts to enable getting predictions by timerange

### DIFF
--- a/cli/src/commands/get/comments.rs
+++ b/cli/src/commands/get/comments.rs
@@ -1,4 +1,4 @@
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
 use chrono::{DateTime, Utc};
 use colored::Colorize;
 use reinfer_client::{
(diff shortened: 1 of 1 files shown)


cli/src/commands/get/comments.rs
```diff
@@ -188,9 +182,9 @@ fn download_comments(
         ))
     };
 
-    if let Some(dataset_identifier) = options.dataset_identifier {
+    if let Some(ref dataset_identifier) = options.dataset_identifier {
         let dataset = client
-            .get_dataset(dataset_identifier)
+            .get_dataset(dataset_identifier.to_owned())
```
mcobzarenco, line 187: is the `ref` and `.to_owned()` needed?, I'm surprised

cli/src/commands/get/comments.rs
```diff
@@ -216,6 +210,7 @@ fn download_comments(
                 &statistics,
                 options.include_predictions,
                 writer,
+                options,
```
mcobzarenco, line 213: Passing `options.include_predictions` and also `options` duplicates the `include_predictions` argument and allows for inconsistent state / ussage where the argument `include_predictions` is different from what's passed in to the `options`.

I think you should either 
 - add ` TimeRange` argument and pass in `options.timerange` (as that's what you need) or 
 - pass in `options` and take `include_predictions` from the object

Given `CommentDownloadOptions` contains a bunch of other stuff you don't need:
```rust
#[derive(Default)]
struct CommentDownloadOptions {
    dataset_identifier: Option<DatasetIdentifier>,
    include_predictions: bool,
    reviewed_only: bool,
    timerange: CommentsIterTimerange,
    show_progress: bool,
}
```
I think you go with the former and just add a `timerange` argument

mcobzarenco (changes requested):
Just some style nits, I'm looking into why the constraint existed
```
"The `dataset` and `from/to-timestamp` options are mutually exclusive."
```

cli/src/commands/get/comments.rs
mcobzarenco: Ah, if you only download review comments, you don't use the `timerage`, see above call to `get_reviewed_comments_in_bulk`

mcobzarenco (approved):
Looks good -- please squash & merge

PR #156, opened 2022-12-02
comments: support fetching auto-thresholded labels

margold:
In order to be able to do an analysis like "X user property by label", I need to know a specific auto-threshold for each predicted label. The idea in this PR is that if a `--model-version` flag is provided to `re get comments --predictions true`, instead of fetching the latest predictions the CLI will fetch predictions from the specified model version with the corresponding auto-threshold information (using the [predict-comments](https://developers.reinfer.io/api/reference/predictions#get-predictions-for-a-pinned-model-by-comment-id) v1 route).

The output is not roundtrippable---I think that's fine? But happy to change if there are better ideas.

### DIFF
--- a/api/src/resources/comment.rs
+++ b/api/src/resources/comment.rs
@@ -89,6 +89,11 @@ pub struct GetAnnotationsResponse {
     pub after: Option<GetLabellingsAfter>,
 }
 
+#[derive(Debug, Clone, Deserialize)]
+pub struct GetPredictionsResponse {
+    pub predictions: Vec<Prediction>,
+}
+
 #[derive(Debug, Clone, Serialize)]
 pub struct UpdateAnnotationsRequest<'a> {
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -367,6 +372,15 @@ pub struct AnnotatedComment {
     pub moon_forms: Option<Vec<MoonForm>>,
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct Prediction {
+    pub uid: Uid,
+    #[serde(skip_serializing_if = "should_skip_serializing_optional_vec")]
+    pub labels: Option<Vec<AutoThresholdLabel>>,
+    #[serde(skip_serializing_if = "should_skip_serializing_optional_vec")]
+    pub entities: Option<Vec<Entity>>,
+}
+
 pub fn get_default_labelling_group(labelling: &Option<Vec<Labelling>>) -> Option<&Labelling> {
     labelling
         .iter()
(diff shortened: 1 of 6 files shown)


api/src/resources/dataset.rs
```diff
@@ -67,6 +67,9 @@ pub struct Id(pub String);
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
 pub struct ModelFamily(pub String);
 
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
+pub struct ModelVersion(pub String);
```
dan-ri, line 71: Shouldn't this be a `u32` or some more structured type, not a string?

dan-ri (approved):
Code looks OK, but will defer to @annesylvie on the approach.

api/src/resources/comment.rs
```diff
@@ -565,6 +588,13 @@ pub struct PredictedLabel {
     pub probability: NotNan<f64>,
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct AutoThresholdLabel {
+    pub name: Vec<String>,
+    pub probability: NotNan<f64>,
+    pub auto_thresholds: Vec<String>,
```
annesylvie, line 595: Isn't that a bunch of numbers instead?

margold: No, it looks like this (coming from the Predict api routes):

```
    {
      "labels": [
        {
          "name": [
            "Margin Call"
          ],
          "probability": 0.9999445676803589,
          "auto_thresholds": [
            "high_recall",
            "balanced",
            "high_precision",
            "sampled_0",
            "sampled_1",
            "sampled_2"
          ]
        },
        {
          "name": [
            "Margin Call",
            "Confirmation"
          ],
          "probability": 0.9999445676803589,
          "auto_thresholds": [
            "high_recall",
            "balanced",
            "high_precision",
            "sampled_0",
            "sampled_1",
            "sampled_2"
          ]
        }
      ]
    }
```
i.e. it shows which thresholds a given predicted label meets.

annesylvie: Oh, didn't realise! I had forgotten how this worked.

margold:
@annesylvie I pushed a change so that the output with `--model-version` is an `AnnotatedComment` (same as the output without `--model-version`) (having updated the definition of labels in `AnnotatedComment` to optionally include the list of auto-thresholds). I had to set a bunch of things to empty as I don't have them in the `predict-comment` response (such as assigned/dismissed), but at least the formats are consistent now.

An alternative way to achieve the desired format with all info would be to call the `predict-comments` API route _on top_ of calling `labellings`, but I think this is not worth it as it would make downloading a large dataset slower.

cli/src/commands/get/comments.rs
```diff
@@ -289,42 +291,69 @@ fn get_comments_from_uids(
...
-                            labelling: prediction.labels,
-                            entities: prediction.entities,
-                        },
-                    );
-                    print_resources_as_json(comments, &mut writer)
-                }
+                            labelling: Some(vec![Labelling {
+                                group: DEFAULT_LABEL_GROUP_NAME.clone(),
+                                assigned: Vec::new(),
+                                dismissed: Vec::new(),
```
annesylvie, line 313: I would add a comment here explaining why you are not populating these fields: because the `predict_comments` endpoint does not return them but we still want the shape of the returned data to be consistent across the different uses of the CLI tool.

Same below for entities.

annesylvie (approved):
LGTM! I didn't realise it was "that" simple to keep the returned jsons in sync, so I think your approach of populating assignments totally makes sense. I do think this way is nicer.

And you are right that calling `get_labellings` to populate annotations would be bad, of course because it would make everything take much longer, but also because we actually don't have an api endpoint to return labellings _as they were_ when a specific model version was created. So populating labellings when `--model-version` is passed as an argument feels just wrong IMO.

PR #157, opened 2022-12-09
api: remove semantic-url-joins

joe-prosser:
The `join` function only joins to the base of a given `url`, this is incompatible with deployments which live in sub directories. This PR seeks to address that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Unreleased
+- Removed semantic url joins to support deployments within a subdirectory
 
 ## Added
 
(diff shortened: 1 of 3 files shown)


tommilligan (approved):
Looks really good, thanks for doing this.

Two points I updated in a review commit:

- Simplified error handling. The error case is just that the base url we're building on isn't valid. So we don't need to bother including path segments or anything like that in the message - it's not possible to have an invalid segment, [they get escaped](https://docs.rs/url/latest/url/struct.PathSegmentsMut.html#method.push)
- Instead of passing a `Vec<T>`, we can just pass a slice `&[T]`. This is more general and one less allocation, plus it's just a bit shorter to write. (slices are to vecs, as string slices are to strings)

PR #163, opened 2023-01-26
datasets: allow passing moon forms through label defs on creation

AWilcke:
Should only be merged after https://github.com/reinfer/platform/pull/16440 adds the required functionality to the api.


api/src/resources/label_def.rs
```diff
@@ -46,3 +51,11 @@ pub struct NewLabelDefPretrained {
...
     pub name: Option<Name>,
 }
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct MoonFormFieldDef {
+    pub name: String,
+    pub kind: String,
+    pub required: bool,
+    pub r#type: String,
+}
```
mcobzarenco, line 61: A comment about the API design -- although not the right place -- but `kind` and `type` are synonyms, I wished we called them something more explicit, e.g. `entity_kind`, `field_type`

Secondly, not sure how I feel about using `r#type` vs `#[serde(rename = "type")` -- fwiw, if we call `type` -> `field_type` (just in the struct for now), we won't be needing `r#`anyway

mcobzarenco (approved):
A naming nit, happy to merge once we rename the `type` struct field

mcobzarenco (approved):

PR #165, opened 2023-02-10
args: Make the -k arg be a plain bool

rcatalinas-reinfer:
This emulates the same -k format used in curl, where the flag doesn't take any parameter, it simpy gets activated if the -k is passed.

I found this when following the platform README, to try the public API, it was confusing that i needed to do `-k true`. Hope this doesn't break anyone's scripts?

### DIFF
--- a/cli/src/args.rs
+++ b/cli/src/args.rs
@@ -36,8 +36,8 @@ pub struct Args {
     /// context, if any.
     pub endpoint: Option<Url>,
 
-    #[structopt(short = "k", long = "accept-invalid-certificates", parse(try_from_str))]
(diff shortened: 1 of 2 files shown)


mcobzarenco:
@rcatalinas-reinfer While inconsistent with curl, the reason this is a bool (and other arguments like this in the CLI) is that it makes it somewhat easier to use in a script (e.g. bash) where you compute a boolean and then just set -- e.g. `--accept_invalid_certificates=$ACCEPT_INVALID`

rcatalinas-reinfer:
Fair enough! I was imagining it could be for that purpose. Gonna close then.

PR #167, opened 2023-02-17
Add `create quota` command

annesylvie:
Add a new `create quota` command, to allow support users to set values for quotas for specific resources.

Example usage:
```bash
re create quota --reinfer-tenant-id abcdefabcdefabcd --quota-kind labels_per_dataset --limit 50
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,11 @@
+# Unreleased
+
+## Added
+
+- `re create quota` to set a quota in a tenant
+
 # v0.12.2
+
 - Rename "triggers" to "streams" following the rename in the API
 - Removed semantic url joins to support deployments within a subdirectory
 - Added functionality to use moon forms both in `LabelDef`s and in `AnnotatedComments`s
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -37,6 +37,7 @@ use crate::resources::{
         CreateProjectRequest, CreateProjectResponse, GetProjectResponse, GetProjectsResponse,
         UpdateProjectRequest, UpdateProjectResponse,
     },
+    quota::{CreateQuota, TenantQuotaKind},
     source::{
         CreateRequest as CreateSourceRequest, CreateResponse as CreateSourceResponse,
         GetAvailableResponse as GetAvailableSourcesResponse, GetResponse as GetSourceResponse,
@@ -48,6 +49,7 @@ use crate::resources::{
         GetResponse as GetStreamsResponse, ResetRequest as StreamResetRequest,
         TagExceptionsRequest as TagStreamExceptionsRequest,
     },
+    tenant_id::TenantId,
     user::GetResponse as GetUserResponse,
     user::{
         CreateRequest as CreateUserRequest, CreateResponse as CreateUserResponse,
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - `re update source` can now update the source's transform tag
 - `re get source` and `re get sources` will show bucket name if exists.
 - `re get comments` can now download predictions within a given timerange
+- `re create quota` to set a quota in a tenant
```
dan-ri, line 19: We discussed whether this should be `create` or `update` given that in some sense there's always a tenant quota, even if it's just the default. But since we're going to add a `delete` method and we don't want to create completely new subcommands, calling this `create` for symmetry seems best.

api/src/resources/quota.rs
```diff
@@ -0,0 +1,127 @@
...
+    Integrations,
+    MailboxesPerIntegration,
+    Triggers,
+    TriggersPerDataset,
+    Users,
+    Alerts,
+    Buckets,
+    Projects,
+    PinnedModels,
+    Unknown(Box<str>),
```
dan-ri, line 26: Remove this variant so this can be a simple enum.

api/src/lib.rs
```diff
@@ -244,6 +245,22 @@ impl Client {
         self.delete(self.endpoints.source_by_id(&source_id)?)
     }
 
+    /// Set a quota
+    pub fn create_quota(
```
dan-ri, line 249: Agreed it would be very useful to have a `get` command as well. Ideally it would show you which quotas were defaults and which had been changed, but this requires some backend changes and can be a future enhancement.

dan-ri (commented):
IIRC this is waiting on an API change to use the tenant name instead of id.

annesylvie:
@dan-ri There might have been some confusion - while I agree that it would be nicer in the future to use the tenant name, I think the PR as is it right now works (even if a bit impractical) and would be valuable to merge in itself.

I will not have the time right now to change the API for using the tenant name instead of ID (maybe I will during our next open day!), so I don't think we should block this PR on the API change.

cli/src/commands/create/quota.rs
```diff
@@ -0,0 +1,37 @@
+use anyhow::{Context, Result};
+use log::info;
+use reinfer_client::{resources::quota::TenantQuotaKind, Client};
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+pub struct CreateQuotaArgs {
+    #[structopt(long = "tenant-id")]
+    /// Tenant ID for which to set the quota
+    tenant_id: String,
```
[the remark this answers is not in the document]
annesylvie: I didn't find a way to do this mutually-exclusive logic at the `structopt` level so I had to handle this somewhere else instead.

cli/src/commands/create/quota.rs
```diff
@@ -0,0 +1,61 @@
...
+    let CreateQuotaArgs {
+        reinfer_tenant_id,
+        uipath_tenant_id,
+        tenant_quota_kind,
+        hard_limit,
+    } = args;
+
+    let tenant_id: TenantId = match (reinfer_tenant_id, uipath_tenant_id) {
+        (Some(tenant_id), None) => TenantId::Reinfer(tenant_id.to_owned()),
+        (None, Some(tenant_id)) => TenantId::UiPath(tenant_id.to_owned()),
```
dan-ri, line 41: Instead of this you should be able to do this with an [argument group](https://github.com/TeXitoi/structopt/issues/104#issuecomment-907318718).

dan-ri (approved):
> I don't think we should block this PR on the API change.

I was thinking the API change would be pretty quick and that switching from one style of CLI arguments to another later might confuse people. Happy to leave the call to you, though.

PR #177, opened 2023-04-12
create: add option to bypass billing

joe-prosser:
Adds an option to allow internal users to bypass billing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
-# v0.13.0
+# Unreleased
+
+## Added
+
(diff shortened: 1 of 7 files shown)


tommilligan (approved):
Updated with a slightly different implementation of how the parameter actually gets added to the request - `reqwest` already exposes a `.query(...)` method, which I think is nicer than this being an endpoint-level concern (this is really a per-request option, not a per-route one).

Other than that, looks good! I haven't tested this still works against the API - could you give it a spin locally before merging?

PR #179, opened 2023-05-22
api: add time series and by label filters to statistics

joe-prosser:
Expands the scope of queries that can be made to the statistics endpoint. Needed for an internal project.

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -12,7 +12,7 @@ use reqwest::{
     header::{self, HeaderMap, HeaderValue},
     IntoUrl, Proxy, Result as ReqwestResult,
 };
-use resources::project::ForceDeleteProject;
+use resources::{dataset::StatisticsRequestParams, project::ForceDeleteProject};
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use std::{cell::Cell, fmt::Display, path::Path};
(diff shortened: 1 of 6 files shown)


api/src/lib.rs
```diff
@@ -610,11 +610,15 @@ impl Client {
...
+    pub fn get_statistics(
+        &self,
+        dataset_name: &DatasetFullName,
+        params: &StatisticsRequestParams,
+    ) -> Result<Statistics> {
         Ok(self
             .post::<_, _, GetStatisticsResponse>(
                 self.endpoints.statistics(dataset_name)?,
-                json!({}),
+                json!(params),
```
tommilligan: While this does what you want, it's better to use `serde_json::to_value` which returns a `Result` (indicating the result is fallible). You can then unwrap/expect/`?` this as you want.

api/src/resources/dataset.rs
```diff
@@ -185,3 +227,54 @@ pub(crate) struct UpdateRequest<'request> {
...
+    use serde_json::json;
+
+    #[test]
+    pub fn test_serialize_statistics_request_params_default() {
+        let params = StatisticsRequestParams {
+            ..Default::default()
+        };
+
+        assert_eq!(
+            json!(params).to_string(),
```
tommilligan: You're doing an extra step here of `struct` -> `serde_json::Value` -> `String`. You can use `serde_json::to_string` to convert straight to a `String`.

Note this has the slightly different behaviour that the fields are serialized in the order they are defined in the `struct` - so I've sorted the struct fields for consistent behaviour.

cli/src/commands/get/comments.rs
```diff
@@ -198,7 +198,12 @@ fn download_comments(
         Ok(get_comments_progress_bar(
             if let Some(dataset_name) = dataset_name {
                 *client
-                    .get_statistics(dataset_name)
+                    .get_statistics(
+                        dataset_name,
+                        &StatisticsRequestParams {
+                            ..Default::default()
+                        },
```
tommilligan: You don't actually need to bother with the additional import for this:
```suggestion
                        &Default::default(),
```

tommilligan (approved):
Pushed a review commit, but all internal implementation things - LGTM!

PR #180, opened 2023-05-23
api: add query endpoint

joe-prosser:
Adds the dataset `query` endpoint to the api client. Needed for an internal project.


tommilligan (approved):
LGTM, nice

PR #186, opened 2023-06-22
api: Add recent as an order option for query api

joe-prosser:
Adds recent as an order option for the query api


jcalero (approved):
LGTM :+1:

PR #188, opened 2023-06-28
api: add user properties filters to query request

joe-prosser:
Allows users to filter on user properties when making query requests


tommilligan (approved):
nice, LGTM :+1:

PR #190, opened 2023-06-29
Add a warning for UiPath cloud users when an operation will charge ai units

joe-prosser:
Adds a warning when users who have permission to bypass billing are not bypassing billing

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -182,7 +182,7 @@ impl Client {
     }
 
     /// Get the base url for the client
-    pub fn get_base_url(&self) -> &Url {
+    pub fn base_url(&self) -> &Url {
         &self.endpoints.base
     }
 
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -181,6 +181,11 @@ impl Client {
         })
     }
 
+    /// Get the base url for the client
+    pub fn get_base_url(&self) -> &Url {
+        &self.endpoints.base
+    }
+
     /// List all visible sources.
     pub fn get_sources(&self) -> Result<Vec<Source>> {
         Ok(self
--- a/api/src/resources/user.rs
+++ b/api/src/resources/user.rs
@@ -1,7 +1,9 @@
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
+use serde_with::{DeserializeFromStr, SerializeDisplay};
 use std::{
     collections::{HashMap, HashSet},
+    fmt::Display,
     str::FromStr,
 };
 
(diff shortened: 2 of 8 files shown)


api/src/resources/user.rs
```diff
@@ -287,8 +324,33 @@ pub struct PostUserResponse {
...
+            GlobalPermission::Debug,
+            GlobalPermission::Demo,
+            GlobalPermission::SubscriptionsRead,
+            GlobalPermission::ArtefactsRead,
+            GlobalPermission::LegacyDialog,
+            GlobalPermission::SupportTenantAdmin,
+            GlobalPermission::SupportUsersWrite,
+            GlobalPermission::TenantAdmin,
+            GlobalPermission::TenantQuotaWrite,
+            GlobalPermission::DeploymentQuotaWrite,
```
tommilligan, line 343: Could you add a boxed str value here as well for the test case

cli/src/commands/create/comments.rs
```diff
@@ -60,9 +63,17 @@ pub struct CreateCommentsArgs {
     #[structopt(short = "n", long = "no-charge")]
     /// Whether to attempt to bypass billing (internal only)
     no_charge: bool,
+
+    #[structopt(short = "w", long = "no-warning")]
+    /// Don't warn when users who have permission to bypass billing are not bypassing billing
+    suppress_no_charge_warning: Option<bool>,
```
tommilligan, line 69: I'd say this flag should be either `-y` (yes) by convention with other linux tooling (see also: apt, kubectl, vogon etc.)

cli/src/commands/create/emails.rs
```diff
@@ -36,9 +39,17 @@ pub struct CreateEmailsArgs {
     #[structopt(short = "n", long = "no-charge")]
     /// Whether to attempt to bypass billing (internal only)
     no_charge: bool,
+
+    #[structopt(short = "w", long = "no-warning")]
+    /// Don't warn when users who have permission to bypass billing are not bypassing billing
+    suppress_no_charge_warning: Option<bool>,
```
tommilligan, line 45: ditto

cli/src/commands/mod.rs
```diff
@@ -1,5 +1,29 @@
...
+        .contains(&reinfer_client::GlobalPermission::Debug)
+    {
+        println!(
+            "\n🚨⚠️ 👉 CAUTION 👈⚠️ 🚨\n\nThe operation you are about to perform will charge AI units\n"
+        );
+        let response = utils::read_from_stdin("To continue, type \"continue\"", None)?;
+
+        if response.to_lowercase() != "continue" {
+            return Err(anyhow!("Billable operation aborted by user"));
+        }
```
tommilligan, line 26: If you want to make your life easier, I suggest adding `dialoguer` and dropping in `Confirm` here (you can give a prompt, then it handles all the retries/input parsing etc. for you): https://docs.rs/dialoguer/latest/dialoguer/struct.Confirm.html

tommilligan (approved):
Looks good - just suggest farming out the low level user interaction to a 3rd party lib

api/src/lib.rs
```diff
@@ -181,6 +181,11 @@ impl Client {
         })
     }
 
+    /// Get the base url for the client
+    pub fn get_base_url(&self) -> &Url {
```
tommilligan, line 185: nit: it's fine to just call this `base_url` (omit the get)

cli/src/commands/create/emails.rs
```diff
@@ -36,9 +39,17 @@ pub struct CreateEmailsArgs {
     #[structopt(short = "n", long = "no-charge")]
     /// Whether to attempt to bypass billing (internal only)
     no_charge: bool,
+
+    #[structopt(short = "y", long = "yes")]
+    /// Consent to ai unit charge. Suppresses confirmation prompt.
+    consent_to_ai_unit_charge: bool,
```
tommilligan, line 45: I would say it's fine to just name this argument `yes`, as in the user has opted to skip all confirmation prompts.

Would be nice for adding future prompts

tommilligan (approved):
LGTM, just a couple of optional nits

PR #193, opened 2023-06-29
cli: include markup for comments

samuelfekete:
Adds markup for comments, as per: https://uipath.atlassian.net/browse/RE-418


annesylvie (approved):
Excellent, thanks a lot!

annesylvie:
Actually, one small thing I forgot: could you add an entry to the Changelog, under "unreleased", mentioning this change please?

PR #198, opened 2023-07-11
quotas: add get quotas command

joe-prosser:
Add `re get quotas` command to list quotas for current tenant


annesylvie (approved):
Thanks!

PR #199, opened 2023-07-11
cli: remove num-format dependency

cristicbz:
Also change quota type to u64, which is what the API actually expects (and what we store in cockroachdb).


joe-prosser (approved):
:+1:

PR #200, opened 2023-07-17
comments: show correct statistics when downloading comments

joe-prosser:
Makes sure the progress bar shows the correct statistics when downloading comments


tommilligan (approved):
LGTM, nice change :+1:

PR #215, opened 2023-08-17
comments: Label filter on get comments

joe-prosser:
Allows you to only download specific comment predictions from a dataset by filtering on label name

### DIFF
--- a/cli/src/commands/get/comments.rs
+++ b/cli/src/commands/get/comments.rs
@@ -234,8 +234,8 @@ fn get_label_attribute_filter(
             "Filtering on label(s):\n- {}",
             label_names
                 .iter()
-                .map(|label_name| label_name.0.clone())
-                .collect::<Vec<String>>()
+                .map(|label_name| label_name.0.as_str())
+                .collect::<Vec<_>>()
                 .join("\n- ")
         );
         Ok(Some(AttributeFilter {
--- a/cli/src/commands/get/comments.rs
+++ b/cli/src/commands/get/comments.rs
@@ -1,10 +1,16 @@
 use anyhow::{bail, Context, Result};
+
 use chrono::{DateTime, Utc};
 use colored::Colorize;
+use log::info;
+use regex::Regex;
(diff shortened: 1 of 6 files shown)


cli/src/commands/get/comments.rs
```diff
@@ -179,14 +212,61 @@ pub fn get_many(client: &Client, args: &GetManyCommentsArgs) -> Result<()> {
...
+        info!("No label names matching the filter '{}'", filter);
+        Ok(None)
+    } else {
+        info!(
+            "Filtering on label(s):\n- {}",
+            label_names
+                .iter()
+                .map(|label_name| label_name.0.clone())
+                .collect::<Vec<String>>()
+                .join("\n- ")
```
tommilligan: Don't need to clone here with
```suggestion
            label_names
                .iter()
                .map(|label_name| label_name.0.as_str())
                .collect::<Vec<_>>()
                .join("\n- ")
```

cli/tests/test_comments.rs
```diff
@@ -246,12 +256,30 @@ fn test_delete_comments_in_range() {
...
+        if result.lines().count() == expected_count {
+            break;
+        } else {
+            thread::sleep(Duration::from_secs(1));
+            result = cli.run(command);
+        }
+    }
+
+    result
+}
```
tommilligan: Yeah agreed with Cristi this is a totally fine approach, we have something similar in some of our python tests here: https://github.com/reinfer/platform/blob/38ff0f6c5e964c41edfa19b84c98fd6261ad417d/exchange_online_sync/reinfer/exchange_online_sync/tests/test_sync.py#L353

You might also prefer to just pull in a crate with an already nice api + exponential backoff like `backoff`: https://lib.rs/crates/backoff

also nit: use `&[&str]` rather than `[&str; 5]` which is a very strict input requirement (fixed length)

tommilligan (approved):
LGTM, just two small nits pushed in a commit

PR #220, opened 2023-08-22
cli: add ability to filter on user properties when getting comments

joe-prosser:
Adds ability to filter on user properties when getting comments, currently only supports string properties with `any_of` filter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## Unreleased
+-Add ability to filter on user properties when getting comments
+
 ## v0.18.1
 - Add comment id to document object in api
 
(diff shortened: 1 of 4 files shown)


tommilligan (commented):
Could you provide a bit more context about `StructExt` and the `from_str` method here? I thought you might be using it to deserialise from a json string input provided by the user, but that doesn't look to be the case?

joe-prosser:
That was my intention yes, so they can either provide the query filter interactively via `--interactive-user-property-filter` or via json string with `--user-property-filter`

tommilligan (approved):
Yep that makes sense - LGTM!

PR #224, opened 2023-08-25
cli: do not print bucket updated timestamp

margold:
It is my understanding that none of the [update-related timestamps on bucket](https://github.com/reinfer/platform/blob/master/store/reinfer/store/bucket/resources.py#L79-L83) are currently being updated (following transition from bucket states to folder states), and we don't have a "most recent folder state update timestamp" field either, so this PR removes the "Updated" column from the CLI's bucket output, as it is now wrong.


joe-prosser (approved):

tommilligan (approved):
LGTM, thanks for tidying this up!

PR #225, opened 2023-09-11
commands: add source statistics to table output

joe-prosser:
Optionally print `num_comments` when getting sources

### DIFF
--- a/cli/src/commands/get/sources.rs
+++ b/cli/src/commands/get/sources.rs
@@ -44,24 +44,22 @@ pub fn get(client: &Client, args: &GetSourcesArgs, printer: &Printer) -> Result<
         .map(|bucket| (bucket.id.clone(), bucket))
         .collect();
 
-    let source_stats: HashMap<_, _> = if *include_stats {
-        sources
-            .iter()
-            .map(|source| {
-                info!("Getting statistics for source {}", source.full_name().0);
-                let stats = client
-                    .get_source_statistics(
-                        &source.full_name(),
-                        &StatisticsRequestParams {
-                            comment_filter: Default::default(),
-                        },
-                    )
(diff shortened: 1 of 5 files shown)


cli/src/commands/get/sources.rs
```diff
@@ -35,6 +44,26 @@ pub fn get(client: &Client, args: &GetSourcesArgs, printer: &Printer) -> Result<
...
+                    .get_source_statistics(
+                        &source.full_name(),
+                        &StatisticsRequestParams {
+                            comment_filter: Default::default(),
+                        },
+                    )
+                    .expect("Could not get statistics for source");
+                (source.id.clone(), stats)
+            })
+            .collect()
```
tommilligan, line 62: Should raise error rather than `expect`ing here, as per previous reviews:
```suggestion
        sources
            .iter()
            .map(|source| {
                info!("Getting statistics for source {}", source.full_name().0);
                let stats = client
                    .get_source_statistics(
                        &source.full_name(),
                        &StatisticsRequestParams {
                            comment_filter: Default::default(),
                        },
                    )
                    .context("Could not get statistics for source")?;
                Ok((source.id.clone(), stats))
            })
            .collect()?
```
or something like this

cli/src/printer.rs
```diff
@@ -182,7 +185,12 @@ impl DisplayTable for PrintableSource {
                     None => "none".dimmed(),
                 },
             },
-            self.source.title
+            self.source.title,
+            if let Some(stats) = &self.stats {
+                stats.num_comments.to_string()
+            } else {
+                "N/A".to_string()
```
tommilligan, line 192: For consistency with above?
```suggestion
                "none".dimmed()
```

tommilligan (approved):
LGTM, two minor comments

PR #229, opened 2023-09-25
cli: rename  to

joe-prosser:
As it actually creates multiple streams


tommilligan (approved):
For reference, it's fairly common to have both words as a subcommand pointing to the same place.

```diff
diff --git a/cli/src/commands/create/mod.rs b/cli/src/commands/create/mod.rs
index 2c7a975..6b8ea13 100644
--- a/cli/src/commands/create/mod.rs
+++ b/cli/src/commands/create/mod.rs
@@ -67,6 +67,10 @@ pub enum CreateArgs {
     #[structopt(name = "stream")]
     /// Create a stream
     Stream(CreateStreamsArgs),
+
+    #[structopt(name = "streams")]
+    /// Create a stream
+    Streams(CreateStreamsArgs),
 }
 
 pub fn run(
@@ -90,6 +94,8 @@ pub fn run(
             stream_exception::create(&client, stream_exception_args, printer)
         }
         CreateArgs::Quota(quota_args) => quota::create(&client, quota_args),
-        CreateArgs::Stream(stream_args) => streams::create(&client, stream_args),
+        CreateArgs::Stream(stream_args) | CreateArgs::Streams(stream_args) => {
+            streams::create(&client, stream_args)
+        }
     }
 }
```

I would say this is better to not break back-compatibility, but up to you. I don't think the burden of supporting two variants it significant.

PR #231, opened 2023-10-09
commands: add get dataset stats

joe-prosser:
Adds flag to get stats for datasets

### DIFF
--- a/api/src/resources/dataset.rs
+++ b/api/src/resources/dataset.rs
@@ -40,6 +40,7 @@ pub struct Dataset {
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
 pub struct DatasetStats {
     pub num_reviewed: NotNan<f64>,
+    pub total_verbatims: NotNan<f64>,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
--- a/api/src/resources/dataset.rs
+++ b/api/src/resources/dataset.rs
@@ -17,7 +17,7 @@ use std::{
     str::FromStr,
 };
 
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
 pub struct Dataset {
     pub id: Id,
     pub name: Name,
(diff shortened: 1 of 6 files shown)


api/src/resources/dataset.rs
```diff
@@ -17,7 +17,7 @@ use std::{
     str::FromStr,
 };
 
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
```
tommilligan, line 20: To avoid having to remove the `Eq`, you should use `ordered_float::NotNan<f64>` which we already use for this purpose.

api/src/resources/dataset.rs
```diff
@@ -34,6 +34,7 @@ pub struct Dataset {
     pub entity_defs: Vec<EntityDef>,
     pub label_defs: Vec<LabelDef>,
     pub label_groups: Vec<LabelGroup>,
+    pub num_reviewed: Option<f64>,
```
tommilligan, line 37: Do you need something like a `#[serde(default)]` here, as `num_reviewed` is not actually returned by the datasets api endpoint?

This is kind of breaking the encapsulation of this type a bit - as it lives in `api`, I would expect it to map to our api types.

I think it would be better practice to have a struct like
```rust
struct DatasetAndStats {
    dataset: Dataset,
    stats: Option<DatasetStats>
}
```

and you can implement `DisplayTable` for that instead. Or, just have a type in the `cli` named `Dataset` that you can convert `From` the `api` Dataset, and add some more stats to.

If you decide not to do the above, I think you need to add a comment documenting that this field is not returned by the datasets api

joe-prosser: Yep, great point. I'll update

cli/src/printer.rs
```diff
@@ -99,11 +100,28 @@ impl DisplayTable for Dataset {
...
+            self.dataset.owner.0.dimmed(),
+            "/".dimmed(),
+            self.dataset.name.0
+        );
+        row![
+            full_name,
+            self.dataset.id.0,
+            self.dataset.updated_at.format("%Y-%m-%d %H:%M:%S"),
+            self.dataset.title,
+            self.stats.num_reviewed
```
tommilligan, line 124: Nice that this is so simple - I was thinking you could delegate through to the `Dataset` implementation, but that doesn't look simple and this is trivial, so happy with just duplicating it.

tommilligan (approved):
LGTM, nice!

PR #232, opened 2023-10-23
commands: filter get users by project and permissions

joe-prosser:
Allows you to filter `get users` by a project and permissions

### DIFF
--- a/cli/src/commands/get/users.rs
+++ b/cli/src/commands/get/users.rs
@@ -1,5 +1,5 @@
-use anyhow::{Context, Result};
-use reinfer_client::{Client, UserIdentifier};
+use anyhow::{bail, Context, Result};
+use reinfer_client::{Client, ProjectName, ProjectPermission, UserIdentifier};
 use structopt::StructOpt;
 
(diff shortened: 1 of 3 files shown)


cli/src/commands/get/users.rs
```diff
@@ -9,24 +9,54 @@ pub struct GetUsersArgs {
...
-    match user {
+    let GetUsersArgs {
+        user,
+        project_name_filter,
+        project_permission_filter,
+    } = args;
+
+    if project_name_filter.is_none() && project_permission_filter.is_some() {
+        bail!("You cannot filter on `permission` without a `project`")
+    }
```
tommilligan, line 31: I would argue you could implement this such that if you didn't pass in a `project`, you got back all users with that permission in any project.

But if that's not the behaviour you want, happy with this.

tommilligan (approved):
Looks good to me

cli/src/commands/get/users.rs
joe-prosser: Yeah good shout - will save that for a future PR. Thanks

PR #233, opened 2023-10-25
parse: add feature to parse and upload msg files

joe-prosser:
Adds a feature to parse unicode msg files and upload them directly to a source. Extracts the minimum information we require from the msg using `cfb` crate


cli/src/commands/parse/msgs.rs
```diff
@@ -0,0 +1,426 @@
...
+            if let Ok(path) = path {
+                if !path.path().extension().is_some_and(|msg| msg == "msg") {
+                    None
+                } else {
+                    Some(path)
+                }
+            } else {
+                None
+            }
+        })
```
tommilligan, line 283: Written slightly strangely - nested `if` normally indicates you can use a combinator, in this case `.ok()`. Also you negate the if statement for some reason, which makes it harder to read?

```suggestion
        .filter_map(|path| {
            let path = path.ok()?;
            if path.path().extension().is_some_and(|msg| msg == "msg") {
                Some(path)
            } else {
                None
            }
        })
```

cli/src/commands/parse/msgs.rs
```diff
@@ -0,0 +1,426 @@
...
+                if documents.len() >= UPLOAD_BATCH_SIZE {
+                    upload_batch_of_documents(
+                        client,
+                        &source,
+                        &documents,
+                        transform_tag,
+                        *no_charge,
+                        &statistics,
+                    )?;
+                    documents.clear();
```
tommilligan, line 321: As you call this more than once (again after the loop), you can easily pull this out into a closure:

```rust
    let send = |documents: &mut Vec<Document>| -> Result<()> {
        upload_batch_of_documents(
            client,
            &source,
            documents,
            transform_tag,
            *no_charge,
            &statistics,
        )?;
        documents.clear();
        Ok(())
    };
```

cli/src/commands/parse/msgs.rs
```diff
@@ -0,0 +1,426 @@
...
+        STREAM_PATH_ATTACHMENT_STORE_PREFIX, attachment_number
+    ))
+}
+
+fn read_attachment(
+    attachment_path: PathBuf,
+    compound_file: &mut CompoundFile<File>,
+) -> Result<AttachmentMetadata> {
+    let mut attachment_name_path = attachment_path.clone();
+    attachment_name_path.push(STREAM_PATH_ATTACHMENT_FILENAME.clone());
```
tommilligan, line 164: nit: don't have to clone here
```suggestion
    attachment_name_path.push(&*STREAM_PATH_ATTACHMENT_FILENAME);
```

`*` dereferences your `Lazy` instance to the inner value, then `&` borrows the `PathBuf` into a `Path` (which is all you need for `push` here)

cli/src/commands/parse/msgs.rs
```diff
@@ -0,0 +1,426 @@
...
+        let mut buffer = Vec::new();
+        stream.read_to_end(&mut buffer)?;
+        buffer
+    };
+
+    Ok(data)
+}
+
+fn read_unicode_stream_to_string(
+    stream_path: PathBuf,
```
tommilligan, line 129: Here and in a few other places you don't actually need a `PathBuf`, just a `Path`, which allows you to remove a bunch of `clone()`s elsewhere.

tommilligan (approved):
Nice work getting this up and running, it looks excellent to me. Honestly I don't think there's anything stopping us from just backporting this code into the api, so we can support `msg` files on the server side there (if we actually want to do that).

Nothing from me - just style/nits which I've pushed review commits for

PR #235, opened 2023-10-26
publish: explicit name for deb package

joe-prosser:
Seems like the default name changed, making it explicit to get around this.


tommilligan (commented):
LGTM, agree this was the right solution

PR #239, opened 2023-11-08
feat(stream-stats): add command for displaying stream validation stats

joe-prosser:
Adds a `get stream-stats` command. The objective of which is to make it easier to understand the impact of upgrading a model version without changing thresholds. The command lets you;

- Get Stream Validation Stats on their own
- Get Stream Validation Stats and compare those stats to a specified model version / dataset and prints a comparison in precision and recall at the same thresholds

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 ## Unreleased
+
 - Fix url used for fetching streams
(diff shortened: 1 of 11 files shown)


tommilligan (approved):
Looks awesome - nice work collecting the label stats in parallel, and in coloring the output for readability. Nothing from me!

PR #241, opened 2023-11-08
add more stream label validation stats

joe-prosser:
Adds some additional data for the stream stats. Namely;
 - What precision would you expect if the recall was the same. What threshold do you need for that?
 - What precision would you expect if the precision was the same. What threshold do you need for that?

### DIFF
--- a/cli/src/commands/get/streams.rs
+++ b/cli/src/commands/get/streams.rs
@@ -103,16 +103,24 @@ pub struct StreamStat {
     recall: NotNan<f64>,
     compare_to_precision: Option<NotNan<f64>>,
     compare_to_recall: Option<NotNan<f64>>,
+    maintain_recall_precision: Option<NotNan<f64>>,
+    maintain_recall_threshold: Option<NotNan<f64>>,
+    maintain_precision_recall: Option<NotNan<f64>>,
+    maintain_precision_threshold: Option<NotNan<f64>>,
 }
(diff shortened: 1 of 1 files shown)


cli/src/commands/get/streams.rs
```diff
@@ -131,46 +139,139 @@ impl DisplayTable for StreamStat {
...
 fn red_if_lower_green_otherwise(test: NotNan<f64>, threshold: NotNan<f64>) -> ColoredString {
     let test_str = format!("{:.3}", test);
 
+    let diff = test - threshold;
+
     match test {
-        test if test < threshold => format!("{test_str} (decrease)").red(),
-        test if test > threshold => format!("{test_str} (increase)").green(),
+        test if test < threshold => format!("{test_str} ({diff:.3})").red(),
+        test if test > threshold => format!("{test_str} (+{diff:.3})").green(),
```
tommilligan, line 173: You can use the printf directive `+` to always add the appropriate string here:

```rust
fn main() {
    for value in [0.0, -0.0, 0.0, -42.3, 42.3] {
        println!("{:+0.3}", value);
    }
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=54b7555880b48aa31f24fd5f3fc499e1

tommilligan (approved):
LGTM, just one nit on the formatting

PR #243, opened 2023-11-16
commands: add get audit events command

joe-prosser:
Adds a command to return audit events

### DIFF
--- a/api/src/resources/audit.rs
+++ b/api/src/resources/audit.rs
@@ -1,0 +1,201 @@
+use chrono::{DateTime, Utc};
+
+use crate::{Continuation, DatasetId, DatasetName, Email, ProjectName, UserId, Username};
+
+use super::{comment::CommentTimestampFilter, project::Id as ProjectId};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct AuditQueryFilter {
+    pub timestamp: CommentTimestampFilter,
+}
+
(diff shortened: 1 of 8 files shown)


api/src/resources/audit.rs
```diff
@@ -0,0 +1,209 @@
...
+
+pub struct QueryResponseIterator {
+    response: AuditQueryResponse,
+    index: usize,
+}
+
+impl Iterator for QueryResponseIterator {
+    type Item = PrintableAuditEvent;
+    fn next(&mut self) -> Option<Self::Item> {
+        let response = self.response.audit_events.get(self.index).map(|event| {
```
tommilligan, line 135: To avoid nesting here, I would return early if you get `None` like:
```suggestion
        let response = self.response.audit_events.get(self.index)?;
        let actor_email = ...
```

api/src/resources/audit.rs
```diff
@@ -0,0 +1,209 @@
...
+            let actor_email = &self
+                .response
+                .get_user(&event.actor_user_id)
+                .unwrap_or_else(|| panic!("Could not find user for id `{}`", event.actor_user_id.0))
+                .email;
```
tommilligan, line 140: If perf is ever an issue here, I would suggest converting the raw `response` into some hashmap lookups for users/datasets etc. at the point you convert it to `QueryResponseIterator`, so you're not repeatedly scanning with `find`.

But happy this works.

Note that I think it is technically possible for us to not return all linked resources (e.g. if the user has been deleted, won't be included).

I'm happy for your cases this is not likely, but if you were pushing this out to other users, I'd want some nicer error handling here.

PR #248, opened 2024-01-08
feat(comments): add messages filters

joe-prosser:
Adds the ability to filter on senders and recipients when running `get comments`

### DIFF
--- a/cli/src/commands/get/comments.rs
+++ b/cli/src/commands/get/comments.rs
@@ -8,8 +8,8 @@ use regex::Regex;
 use reinfer_client::{
     resources::{
         comment::{
-            CommentTimestampFilter, ReviewedFilterEnum, UserPropertiesFilter,
-            UserPropertyFilterKind,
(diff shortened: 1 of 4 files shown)


cli/src/commands/get/comments.rs
```diff
@@ -328,6 +358,7 @@ pub fn get_many(client: &Client, args: &GetManyCommentsArgs) -> Result<()> {
         show_progress: !no_progress,
         label_attribute_filter,
         user_properties_filter,
+        messages_filter: Some(messages_filter),
```
tommilligan, line 361: Should this be `None` if both senders and recipients are None? Or does it not really matter api/perf wise

tommilligan (approved):
LGTM

cli/src/commands/get/comments.rs
joe-prosser: It doesn't matter api wise :)

PR #249, opened 2024-01-17
 fix(api): remove required and type from moon label def

joe-prosser:


AWilcke (approved):
Thanks for cleaning this up! :broom:

PR #254, opened 2024-01-30
feat(integrations): Add integration commands

joe-prosser:


cli/src/commands/create/integrations.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+fn overwrite_integration(
+    client: &Client,
+    name: &IntegrationFullName,
+    new_integration: &NewIntegration,
+    old_integration: &Integration,
+) -> Result<()> {
+    let old_integration: NewIntegration =
+        serde_json::from_str(&serde_json::to_string(old_integration)?)?;
```
tommilligan, line 64: Not great, not terrible :shrug: you could have implemented `From` but this does the job.

tommilligan (approved):

PR #268, opened 2024-04-04
fix(commands): fix issue when getting streams with multiple filters on a single user properties

joe-prosser:


tommilligan (approved):
LGTM, nice fix

PR #269, opened 2024-04-08
fix(parse): case insensitive file name match

joe-prosser:
Resolves: https://github.com/reinfer/cli/issues/263


tommilligan (approved):
I mean - I guess, but as this is nonstandard linux behaviour (files _are_ case sensitive in linux), can we make this something like `get_files_in_directory_case_insensitive` to make clear the behaviour?

That or pass an arg like `case_insensitive: bool` to opt in to case insensitive searching.

PR #274, opened 2024-05-13
feat(quotas): show usage on get quotas

joe-prosser:
show usage on `get quotas`

### DIFF
--- a/api/src/resources/quota.rs
+++ b/api/src/resources/quota.rs
@@ -1,7 +1,7 @@
 use crate::error::{Error, Result};
 use serde::{Deserialize, Serialize};
 use serde_with::{DeserializeFromStr, SerializeDisplay};
-use std::{fmt::Display, str::FromStr};
+use std::{fmt::Display, str::FromStr, u64};
 
 #[derive(Debug, Clone, SerializeDisplay, DeserializeFromStr, PartialEq, Eq, Hash, Copy)]
 pub enum TenantQuotaKind {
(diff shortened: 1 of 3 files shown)


api/src/resources/quota.rs
```diff
@@ -1,7 +1,7 @@
 use crate::error::{Error, Result};
 use serde::{Deserialize, Serialize};
 use serde_with::{DeserializeFromStr, SerializeDisplay};
-use std::{fmt::Display, str::FromStr};
+use std::{fmt::Display, str::FromStr, u64};
```
tommilligan, line 4: shouldn't need to import u64?

cli/src/printer.rs
```diff
@@ -98,11 +98,19 @@ impl DisplayTable for Bucket {
...
     fn to_table_row(&self) -> Row {
-        row![self.quota_kind, Thousands(self.hard_limit)]
+        row![
+            self.quota_kind,
+            Thousands(self.hard_limit),
+            Thousands(self.current_max_usage),
+            format!(
+                "{:.0}%",
+                (self.current_max_usage as f64 / self.hard_limit as f64) * 100.0
+            )
```
tommilligan, line 112: You probably want edge case handling for if `self.hard_limit == 0`, otherwise you'll get `NaN%` here, in the divide by zero case

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d2ddc482f0429345cbef5436efee6bd9

tommilligan (approved):
LG after couple of nits

PR #281, opened 2024-06-28
fix(stats): update api schema for bucket statistics

alexbrt:
The bucket count is now returned as either an exact count or as a lower bound to avoid expensive aggregations for buckets containing more than 100K emails.


annesylvie (approved):
Nicely done! Tested locally as well, and it works perfectly. :sparkles:

PR #284, opened 2024-07-04
feat(commands): add general fields to create dataset command

joe-prosser:


api/src/resources/dataset.rs
```diff
@@ -244,6 +244,9 @@ pub struct NewDataset<'request> {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub entity_defs: Option<&'request [NewEntityDef]>,
 
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub general_fields: Option<&'request [NewGeneralFieldDef]>,
+
```
AWilcke, line 249: Could you also add it to `Dataset`? such that we can easily roundtrip them :pray:

api/src/resources/entity_def.rs
```diff
@@ -23,3 +23,17 @@ pub struct NewEntityDef {
...
     pub trainable: bool,
 }
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct NewGeneralFieldDef {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub field_type_id: Option<String>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub title: Option<String>,
```
AWilcke, line 33: `title`s for General Fields are deprecated, we're only operating on `api_names` going forward. If you look at what the webapp does, it's actually sending a random alphanumeric as the title and there is no way to set the title from the ui

image

AWilcke (changes requested):
Thanks for handling this - CLI hero :superhero:

AWilcke (approved):
Amazing

PR #288, opened 2024-07-23
fix(http): increase default timeout to 120s

joe-prosser:
For long running calls like stream results

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -36,7 +36,7 @@ use resources::{
 };
 use serde::{Deserialize, Serialize};
 use serde_json::json;
-use std::{cell::Cell, fmt::Display, path::Path};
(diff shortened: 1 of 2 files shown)


api/src/lib.rs
```diff
@@ -1844,10 +1844,14 @@ impl Endpoints {
     }
 }
 
+const DEFAULT_HTTP_TIMEOUT: Lazy<Option<Duration>> = Lazy::new(|| Some(Duration::from_secs(120)));
```
dan-ri, line 1847: How about just making this
```rust
const DEFAULT_HTTP_TIMEOUT_SECONDS: u32 = 120
```
and then constructing the `Duration` in `.timeout()`?

dan-ri (approved):

PR #289, opened 2024-07-24
feat(commands): add lossy flag when creating annotations

joe-prosser:
Adding a flag to attempt to resume processing when operations fail. Starting with annotations but could later be used for comments also

### DIFF
--- a/cli/src/commands/create/annotations.rs
+++ b/cli/src/commands/create/annotations.rs
@@ -1,7 +1,4 @@
-use crate::{
-    print_error_as_warning,
-    progress::{Options as ProgressOptions, Progress},
-};
+use crate::progress::{Options as ProgressOptions, Progress};
 use anyhow::{Context, Result};
 use colored::Colorize;
 use log::info;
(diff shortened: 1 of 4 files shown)


cli/src/commands/create/annotations.rs
```diff
@@ -337,14 +363,29 @@ impl Statistics {
...
+        format!(
+            "{} {}{}",
+            num_annotations,
+            "annotations".dimmed(),
+            if num_failed_annotations > 0 {
+                format!(" {} {}", num_failed_annotations, "skipped".dimmed())
+            } else {
+                "".to_string()
+            }
+        ),
```
dan-ri, line 388: The nested `format!` is a bit cheeky. How about:
```rust
let failed_annotations_string = if num_failed_annotations > 0 {
    format!(" {num_failed_annotations} {}", "skipped".dimmed())
} else {
    String::new()
};
```
Similarly below.

cli/src/main.rs
```diff
@@ -203,15 +203,26 @@ fn find_configuration(args: &Args) -> Result<PathBuf> {
...
+        warn!(" |- {}", cause);
+    }
+}
+
+pub fn print_error(error: &anyhow::Error) {
+    error!("An error occurred:");
+    for cause in error.chain() {
+        error!(" |- {}", cause);
+    }
+}
```
dan-ri, line 218: We shouldn't need two functions. How about just
```rust
pub fn print_error(error: &anyhow::Error) {
    for cause in error.chain() {
        error!(" |- {}", cause);
    }
}
```
and leave the call to `error!()`before it in `main()` below. You can add the call to `warn!()` in `upload_batch_of_annotations()`, but I'm not sure logging in that case is necessary.

dan-ri (approved):

PR #290, opened 2024-07-24
fix(commands): only create attachments dir when needed

joe-prosser:


cli/src/commands/get/comments.rs
```diff
@@ -360,7 +365,7 @@ pub fn get_many(client: &Client, args: &GetManyCommentsArgs) -> Result<()> {
     let OutputLocations {
         jsonl_file,
         attachments_dir,
-    } = get_output_locations(path)?;
+    } = get_output_locations(path, include_attachment_content.is_some_and(|v| v))?;
```
tommilligan, line 368:
```suggestion
    } = get_output_locations(path, include_attachment_content.unwrap_or_default())?;
```

tommilligan (approved):

PR #291, opened 2024-07-25
feat(commands): add resume on error for comments and emails

joe-prosser:
Adds a `--resume-on-error` flag for uploading comments and emails. When the api returns an error the CLI will split a particular batch into individual requests, attempt each one individually and skip any individual request that fail

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -36,7 +36,13 @@ use resources::{
 };
 use serde::{Deserialize, Serialize};
 use serde_json::json;
-use std::{cell::Cell, fmt::Display, io::Read, path::Path, time::Duration};
+use std::{
+    cell::Cell,
+    fmt::{Debug, Display},
+    io::Read,
+    path::Path,
+    time::Duration,
+};
 use url::Url;
 
 use crate::resources::{
@@ -136,6 +142,38 @@ pub use crate::{
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Token(pub String);
+
+pub trait SplittableRequest {
+    fn split(self) -> impl Iterator<Item = Self>
+    where
+        Self: Sized;
+
+    fn count(&self) -> usize;
+}
+
+pub struct SplitableRequestResponse<ResponseT>
+where
+    for<'de> ResponseT: Deserialize<'de> + ReducibleResponse,
+{
+    pub response: ResponseT,
+    pub num_failed: usize,
+}
+
+pub trait ReducibleResponse {
+    fn merge(self, _b: Self) -> Self
+    where
+        Self: std::default::Default,
+    {
+        Default::default()
+    }
+
+    fn empty() -> Self
+    where
+        Self: std::default::Default,
+    {
+        Default::default()
+    }
+}
 
 pub struct Config {
     pub endpoint: Url,
@@ -434,13 +472,13 @@ impl Client {
         integration: &NewIntegration,
     ) -> Result<PostIntegrationResponse> {
         self.request(
-            Method::POST,
-            self.endpoints.integration(name)?,
-            Some(PostIntegrationRequest {
+            &Method::POST,
+            &self.endpoints.integration(name)?,
+            &Some(PostIntegrationRequest {
                 integration: integration.clone(),
             }),
-            None::<()>,
-            Retry::No,
+            &None::<()>,
+            &Retry::No,
         )
     }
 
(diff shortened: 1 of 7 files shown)


api/src/resources/comment.rs
```diff
@@ -170,16 +170,63 @@ pub struct CommentsIterPage {
...
+#[derive(Default, Debug, Clone, Deserialize)]
 pub struct PutCommentsResponse;
 
+impl ReducableResponse for PutCommentsResponse {
+    fn empty() -> Self {
+        Self {}
+    }
+    fn merge(_a: Self, _b: Self) -> Self {
+        Self {}
+    }
```
dan-ri, line 205: You can supply a [default implementation](https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations) for trait methods in the trait itself that will be used if there's no implementation for the type. This should allow you to avoid repeating these trivial implementations (you'll need a `Default` [trait bound](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-supertraits-to-require-one-traits-functionality-within-another-trait)). 
```

File: api/src/lib.rs [hunk 2]
```diff
@@ -1214,6 +1298,73 @@ impl Client {
...
+    }
+
+    fn should_split(error: &Error) -> bool {
+        if let Error::Api { status_code, .. } = error {
+            *status_code == reqwest::StatusCode::UNPROCESSABLE_ENTITY
+                || *status_code == reqwest::StatusCode::BAD_REQUEST
+        } else {
+            false
+        }
+    }
```
dan-ri, line 1366: You can define this within `splittable_request`, as it's a helper private to the method.

api/src/lib.rs
```diff
@@ -137,6 +143,27 @@ pub use crate::{
...
+
+pub struct SplitableRequestResponse<ResponseT>
+where
+    for<'de> ResponseT: Deserialize<'de> + ReducableResponse,
+{
+    pub response: ResponseT,
+    pub num_failed: usize,
+}
+
+pub trait ReducableResponse {
```
dan-ri, line 162:
```suggestion
pub trait ReducibleResponse {
```

api/src/lib.rs
```diff
@@ -1214,6 +1298,73 @@ impl Client {
...
+
+                    body.split().iter().for_each(|request| {
+                        match self.request(
+                            method.clone(),
+                            url.clone(),
+                            Some(request),
+                            query.clone(),
+                            retry.clone(),
+                        ) {
+                            Ok(r) => response = SuccessT::merge(response.clone(), r),
```
dan-ri, line 1342: Hmm, this mutation and cloning doesn't look quite right. Conceptually we should be able to [`reduce`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.reduce) the successful responses, although we need to filter out and handle the failures somehow.

api/src/resources/comment.rs
```diff
@@ -189,6 +236,36 @@ pub struct SyncCommentsResponse {
...
+        let Self {
+            new: b_new,
+            updated: b_updated,
+            unchanged: b_unchanged,
+        } = b;
+
+        Self {
+            new: a_new + b_new,
+            updated: a_updated + b_updated,
+            unchanged: a_unchanged + b_unchanged,
```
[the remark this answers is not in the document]
joe-prosser: I added the let binding so that we would get a compiler error if new fields were added (to remind us to make any necessary changes here)

api/src/resources/comment.rs
joe-prosser: I chose not to use the `Default` trait initially as it doesn't make sense to have `Default` for these responses unless you're using the `ReducableResponse` trait. Don't have strong opinions here though

dan-ri: In that case you can make your own version of `Default` called `ReducibleDefault` or whatever.

api/src/resources/comment.rs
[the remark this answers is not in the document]
dan-ri: Rather than do that, you can just put a single statement like this at the top:
```rust
// Ensure new fields are added here:
let Self { new: _, updated: _, unchanged: _ } = a;
```

api/src/resources/comment.rs
joe-prosser: Implemented with `Default` for the sake of simplicity :)

api/src/resources/comment.rs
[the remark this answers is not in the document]
joe-prosser: Thanks - just pushed a slightly different implementation

api/src/lib.rs
```diff
@@ -1214,22 +1294,81 @@ impl Client {
...
+            Ok(response) => Ok(SplitableRequestResponse {
+                response,
+                num_failed: 0,
+            }),
+            Err(error) if should_split(&error) => {
+                let mut num_failed = 0;
+
+                let response = SuccessT::reduce(
+                    body.split()
+                        .iter()
```
dan-ri, line 1334: This is looking much better, but we should be able to do better still as this is a standard fold pattern, if you switch back to the `merge` trait method:
```rust
let response = body
    .split()
    .iter()
    .filter_map(|request| {
        match self.request(&method, &url, &Some(request), &query, &retry) {
            Ok(response) => Some(response),
            Err(_) => {
                num_failed += 1;
                None
            }
        }
    })
    .fold(SuccessT::empty(), |merged, next: SuccessT| merged.merge(next));
```

api/src/lib.rs
```diff
@@ -137,6 +143,35 @@ pub use crate::{
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Token(pub String);
 
+pub trait SplittableRequest {
+    fn split(self) -> Vec<Self>
```
dan-ri, line 147: Also, you should be able to define `split` as returning an `impl Iterator` instead of a `Vec` and remove the `collect()` at the end of the implementations.

dan-ri (approved):

PR #297, opened 2024-08-05
chore(api): retry put comments requests

joe-prosser:
May result on 409 on retry if the put actually worked, but adding anyway as (anecdotally) most of the time the connection doesn't make it to us


tommilligan (approved):
Fine, but could you add a comment about why you changed this to `Retry::Yes` given the potential for a conflict

PR #298, opened 2024-08-05
feat(commands): add only with attachments filter on get comments

joe-prosser:


cli/src/commands/get/comments.rs
```diff
@@ -122,6 +122,10 @@ pub struct GetManyCommentsArgs {
     #[structopt(long = "attachments")]
     /// Save attachment content for each comment
     include_attachment_content: Option<bool>,
+
+    #[structopt(long = "--only-with-attachments")]
+    /// Save attachment content for each comment
```
tommilligan, line 127: Incorrect docstring

tommilligan (approved):

PR #299, opened 2024-08-05
fix(cli): collect comments before downloading attachments

joe-prosser:
Fixes bug where comments would not be written when downloading attachments


cli/src/commands/get/comments.rs
```diff
@@ -608,12 +608,6 @@ fn download_comments(
                 let page = page.context("Operation to get comments has failed.")?;
                 statistics.add_comments(page.len());
 
-                if let Some(attachments_dir) = &options.attachments_dir {
```
joe-prosser, line 611: We don't allow downloading attachments when you don't have a `dataset`

PR #320, opened 2024-10-01
chore(api): round trip field id

joe-prosser:


api/src/resources/comment.rs
```diff
@@ -875,10 +875,14 @@ pub struct NewEntityWithSpan {
...
 }
 
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Eq)]
+pub struct FieldId(pub String);
+
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Eq)]
 pub struct NewEntityWithSpans {
     pub name: EntityName,
     pub formatted_value: String,
+    pub field_id: FieldId,
```
AWilcke, line 885: We should also add it to `NewEntityWithSpan` for the legacy cases no? And what about the non-new one, like `Entity`? actually I see that `Entity` doesn't support the legacy `span` case either...

AWilcke: > I think I'll address this in a separate PR

yeah that's totally fair, it's just a bit sad to see the state we left the cli in :grimacing: thanks for cleaning up the mess

AWilcke (approved):
Thanks :broom:

PR #335, opened 2025-02-05
feat(cli): add ability to parse pst files

joe-prosser:
Adds a command called `re parse pst` which allows users to parse and upload data from PST files.


scripts/publish-install
```diff
@@ -5,16 +5,28 @@ set -ex
...
     pip install toml
-    brew install mingw-w64
+    brew install                  \
+      mingw-w64                   \
+      autoconf                    \
+      automake                    \
+      gettext                     \
+      libtool                     \
+      pkgconfig
     ;;
```
tommilligan, line 30: windows build next? :wink:

cli/src/commands/parse/pst.rs
```diff
@@ -0,0 +1,1275 @@
...
+        errors
+            .into_iter()
+            .collect::<HashSet<String>>()
+            .iter()
+            .join("\n")
+    }
+}
+
+#[link(name = "pff")]
+extern "C" {
```
tommilligan, line 102: The low level C wrapper should/can be moved out to a separate module called something like `pff-sys`. Basically keep all the low level C-repr details in another file, which basically defines the `libpff` C interface

See this thread: https://users.rust-lang.org/t/should-i-follow-the-sys-crate-convention-for-wrapping-non-system-c-libraries/65150

cli/src/commands/parse/pst.rs
```diff
@@ -0,0 +1,1275 @@
...
+            0 => Ok(LibPffAttachmentType::Undefined),
+            x if x == 'd' as isize => Ok(LibPffAttachmentType::Data),
+            x if x == 'i' as isize => Ok(LibPffAttachmentType::Item),
+            x if x == 'r' as isize => Ok(LibPffAttachmentType::Reference),
+            _ => Err(anyhow!("Unknown Attachment type {}", value)),
+        }
+    }
+}
+
+struct PstAttachment {
```
tommilligan, line 274: These structs that you're defining as the nice api is fine to stay in this file

cli/src/commands/parse/pst.rs
```diff
@@ -0,0 +1,1275 @@
...
+
+#[repr(C)]
+#[derive(Debug, Clone)]
+struct LibPffFile {
+    _unused: c_int,
+}
+
+#[repr(C)]
+#[derive(Debug, Clone)]
+struct LibcError {
```
tommilligan, line 54: I took a look into the source code as I expected this to be `LibPffError` - I agree this is correct, I was wondering if you could use the `libc` crate here, but I don't know enough to comment

Basically, this is mildly surprising but looks fine

cli/src/commands/parse/pst.rs
```diff
@@ -0,0 +1,1275 @@
...
+            }
+
+            if !message_ptr.is_aligned() {
+                errors.push("Unknown Error: message_ptr misaligned".to_string());
+                continue;
+            }
+
+            let message = unsafe { std::ffi::CStr::from_ptr(*message_ptr) }
+                .to_string_lossy()
+                .into_owned();
```
tommilligan, line 88: Generally for `unsafe` blocks, you should leave a comment explaining why it's safe (or assumptions you're making/invariants the compiler can't enforce). This is similar to `expect` or `unwrap` in safe code.

So here I'd say something like

```rust
// Assumes that the pointer returned from `offset` is valid
let message_ptr = unsafe { self.messages.offset(i as isize) };
...
// Assumes the pointer returned from libpff is a valid CString
let message = unsafe { std::ffi::CStr::from_ptr(*message_ptr) }
```
```

File: cli/src/commands/parse/pst.rs [hunk 6]
```diff
@@ -0,0 +1,1275 @@
...
+
+#[derive(Clone)]
+struct PstMessage {
+    pub inner: *mut LibPffItem,
+    pub folder: FolderPath,
+}
+
+impl PstMessage {
+    // Useful for debugging
+    fn _get_message_id(&self) -> Result<String> {
```
tommilligan, line 650: Generally Rust doesn't use a leading underscore for privacy - if you want to keep this private:
- Split `PstMessage` into it's own module, in which case this will be private an inacessible
- Call it `get_message_id_for_debug` or something
- Just call it `get_message_id`. Is there a reason you wouldn't want this used in other code?

cli/src/commands/parse/pst.rs
```diff
@@ -0,0 +1,1275 @@
...
+            .context(format!("{} header is missing", header))?;
+
+        if value.is_empty() {
+            Err(anyhow!("{} header is empty", header))
+        } else {
+            Ok(value)
+        }
+    }
+
+    pub fn into_new_email(self, mailbox: Mailbox) -> Result<NewEmail> {
```
tommilligan, line 672: I think this method could be in a module completely by itself, as it should just use the public API of `PstMessage` and output `NewEmail`.

This would isolate all the business logic of converting a pst message to an email from the internals you've written.

cli/src/commands/parse/pst.rs
```diff
@@ -0,0 +1,1275 @@
...
+            })
+            .collect::<Vec<NewEmail>>();
+
+        let expected_emails: Vec<NewEmail> =
+            serde_json::from_str(include_str!("../../../tests/samples/test.pst.json"))
+                .expect("Could not read expected emails");
+
+        assert_eq!(expected_emails, emails);
+    }
+}
```
tommilligan, line 1275: I'm a little suprised that this test (outputting data in JSONL) is not what the pst-extract type command does - generates you a file that you can then just upload using existing cli functionality.

This would separate the PST conversion and upload tasks, which I imagine probably both have their own kind of failure cases that it would be annoying to mix?

tommilligan (approved):
Looks really good overall. My main comment is about splitting stuff up and keeping it modular. Notably:
- A module for `pff-sys`
- A module for the nice `pff` wrapper
- A module for the business logic of `PffMessage -> RawEmail`
- A module for cli args, IO, upload to api/dump to file etc.

This would help keep the code maintainable (and reviewable)

cli/src/commands/parse/pst.rs
joe-prosser: Agree that it causes weird errors to mix however it follows the same pattern as `re parse eml` / `re parse msg` so prefer to be consistent with that

tommilligan: That's fair enough

PR #363, opened 2025-04-08
feat(commands): add package commands

joe-prosser:
Adds `re package download` and `re package upload` where a `package` is a zip containing all the dependencies for an ixp project (dataset def, source def, annotations, comments, documents etc...) allowing users to move around stuff with minimal fuss. 

One day I'll expand the scope of this to include CM projects.

### DIFF
--- a/cli/src/commands/package/download.rs
+++ b/cli/src/commands/package/download.rs
@@ -45,7 +45,7 @@ pub struct DownloadPackageArgs {
     batch_size: usize,
 
     #[structopt(long = "max-attachment-memory-mb", default_value = "256")]
-    /// Maximum number of kb to use to store attachments in memory
+    /// Maximum number of mb to use to store attachments in memory
     max_attachment_memory_mb: u64,
 }
 
@@ -197,7 +197,7 @@ fn package_batch_of_documents(
     });
 
     if let Ok(error) = error_receiver.try_recv() {
-        return Err(anyhow!("Failed to upload document {}", error));
+        return Err(anyhow!("Failed to download document {}", error));
     };
 
     drop(document_sender);
--- a/cli/src/commands/package/download.rs
+++ b/cli/src/commands/package/download.rs
@@ -1,0 +1,368 @@
+use std::{
+    fs::remove_file,
+    path::PathBuf,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        mpsc::channel,
+        Arc,
+    },
+};
+
+use anyhow::{anyhow, Context, Result};
+use reinfer_client::{
+    resources::{
+        attachments::AttachmentMetadata,
+        dataset::{DatasetFlag, QueryRequestParams},
+    },
+    Client, CommentFilter, CommentId, Dataset, DatasetFullName, DatasetIdentifier, DatasetName,
+    HasAnnotations, SourceId,
+};
+use scoped_threadpool::Pool;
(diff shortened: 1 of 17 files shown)


cli/src/commands/package/upload.rs
```diff
@@ -0,0 +1,515 @@
...
+
+    #[structopt(long = "dataset-creation-timeout", default_value = "30")]
+    /// Maximum number of seconds to wait for dataset to be created
+    dataset_creation_timeout: i64,
+}
+
+fn wait_for_dataset_to_exist(dataset: &Dataset, client: &Client, timeout_s: i64) -> Result<()> {
+    let start_time = Utc::now();
+
+    while (start_time - Utc::now()).num_seconds() <= timeout_s {
```
tommilligan, line 53: You should be using the `Instant` type for this kind of call: https://doc.rust-lang.org/std/time/struct.Instant.html

Generally for anything that's performance or timing related that doesn't relate to human times, these are more reliable and faster than anything to do with human time (system clocks can go backwards etc.). Also simpler!

```rust
    let start = Instant::now();
    while start.elapsed().as_secs() <= timeout_s {
```

cli/src/commands/package/upload.rs
```diff
@@ -0,0 +1,515 @@
...
+    }
+
+    Err(anyhow!("Timeout waiting for dataset to be created"))
+}
+
+fn create_dataset(
+    name: DatasetName,
+    label_defs: Vec<LabelDef>,
+    client: &Client,
+    timeout_s: i64,
```
tommilligan, line 75: Timeout should be `u32`, i.e. non-negative

cli/src/commands/package/upload.rs
```diff
@@ -0,0 +1,515 @@
...
+
+#[derive(Debug)]
+pub struct Statistics {
+    document_reads: AtomicUsize,
+    document_uploads: AtomicUsize,
+    annotations: AtomicUsize,
+    failed_annotations: AtomicUsize,
+    failed_reads: AtomicUsize,
+    failed_uploads: AtomicUsize,
+}
```
tommilligan, line 402: I feel like a lot of your stats could be shared with up/download with some generic naming like

```suggestion
#[derive(Debug)]
pub struct Statistics {
    document_package_count: AtomicUsize,
    document_api_count: AtomicUsize,
    annotations: AtomicUsize,
    failed_annotations: AtomicUsize,
    failed_package_count: AtomicUsize,
    failed_api_count: AtomicUsize,
}
```
as these modules are literally mirrors of each other.

But the names do get less readable, so not fussed.

tommilligan (approved):
Amazing work as always

Glad to see we're adopting `.zip` as well as `.json` and `.proto` into our family of file formats :grin:

cli/src/commands/package/download.rs
```diff
@@ -0,0 +1,368 @@
...
+#[allow(clippy::too_many_arguments)]
+fn package_source_contents(
+    dataset: &DatasetFullName,
+    source_id: &SourceId,
+    batch_size: usize,
+    package_writer: &mut PackageWriter,
+    client: &Client,
+    statistics: &Arc<Statistics>,
+    max_attachment_memory_mb: &u64,
+    pool: &mut Pool,
```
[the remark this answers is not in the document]
joe-prosser: Yeah - just get's passed around and used everywhere

PR #447, opened 2026-06-09
feat(cli): add `re prune` to back up and delete old CM data across datasets

joe-prosser:
## What
Adds `re prune`: it deletes two kinds of old data across a set of datasets —
**comments** (from the datasets' sources) and **emails** (from those sources'
buckets) — older than a cutoff, writing a verified, write-once backup to disk
first so the run is recoverable.

```
re prune --datasets D1,D2 --older-than-days N --backup-dir DIR \
         [--mailbox M] [--include-annotated] [--dry-run] [--yes]
```

It backs up both the comments and the emails, verifies every backup file against
a checksum, and only then deletes — sourcing every deleted id from the verified
backup. Backups are written in `re create comments` / `re create emails` format,
so a run restores end-to-end. Annotated comments are kept by default;
`--dry-run` backs up without deleting.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cli/src/commands/prune.rs
```diff
@@ -0,0 +1,1570 @@
...
+            on_batch(&batch)?;
+            batch.clear();
+        }
+        Ok(())
+    })?;
+    if !batch.is_empty() {
+        on_batch(&batch)?;
+    }
+
+    check_integrity(path, expected, count, crc32)
```
tommilligan: Here, your `check_integrity` is _after_ you've actioned each batch - not before. So theoretically, I guess your ids could be corrupted in the file, you action deleting them an only realise the file was corrupt afterwards.

I would have expected an integrity check to be the _first_ thing you do before any actions.

tommilligan: I see in the comments there are some notes about "this function should be called first with a no-op" - I'd say it's easier to just have hash verification as a completely separate function distinct from this batch-style processing, unless there's a reason to combine them.

tommilligan (approved):
LGTM - I like the backup, and then read from backup style for ensuring we have the data locally before actioning.

I was a bit confused about the hash _after_ action pattern (see comments inline), but think as currently used it's fine; just potentially a bit of a footgun for future refactorings.

